PR #52, opened 2022-10-09
Refactor

jubeless:


orchestrator/jobsplanner.go
```diff
@@ -3,11 +3,11 @@ package orchestrator
 import (
 	"context"
 	"fmt"
+	"github.com/streamingfast/substreams/pipeline/store"
```
abourget, line 6: What’s the difference between `pipeline/store/` and `state/` ?

I understand that `state/` should have long been renamed to `store` .. which would make this new package irrelevant.

Let's say we rename `state` to `store`, then we could have the Accessor concept in `state`. More on that later.

pipeline/store/accessor.go
```diff
@@ -0,0 +1,177 @@
...
+type Accessor struct {
+	baseStore           dstore.Store
+	storeMap            map[string]*state.Store
+	SaveInterval        uint64
+	maxBoundaryBlockNum uint64
+	nextSaveBoundary    uint64
+	logger              *zap.Logger
+}
+
+func NewAccessor(saveInterval, minBoundaryBlock, maxBoundaryBlockNum uint64, baseStore dstore.Store, logger *zap.Logger) *Accessor {
```
abourget, line 26: The accessor now has too many responsibilies.

We had discussed a `StoreMap` which would simply pass the map around, and something like a `store.MapGetter` or something, that would allow downstream to simply get the stores and not assign them.

Accessor has grown to be much more.

I remember even thinking the concept of a Store holding its InitialBlock was already too much. A Store ought not to know how it's being used. Perhaps the concept of a `PartialStore` could hold such a number.. with the `Roll()` method.. In normal linear processing, that number is not useful and is just confusing. I remember myself being confused by it.

So a new idea would be to embed the Store in a PartialStore, and use PartialStore where we expect a partial store (in the orchestrated stuff), but not everywhere. The Squasher for example, works with both types, and having clarity as to which is which would be really useful.

pipeline/store/accessor.go
```diff
@@ -0,0 +1,177 @@
...
+	s.storeMap[storeModule.Name] = store
+	return nil
+}
+
+func (s *Accessor) Get(storeName string) (*state.Store, error) {
+	if store, found := s.storeMap[storeName]; found {
+		return store, nil
+	}
+	return nil, ErrStoreNotFound
+}
```
abourget, line 63: initially, the Accessor was just this, wasn't it? Did you find it necessary to add all the rest ?

pipeline/store/accessor.go
```diff
@@ -0,0 +1,177 @@
...
+		storeModule.Name,
+		s.SaveInterval,
+		storeModule.InitialBlock,
+		// gets thme for store
+		hash,
+		storeModule.GetKindStore().UpdatePolicy,
+		storeModule.GetKindStore().ValueType,
+		s.baseStore,
+		s.logger,
+	)
```
abourget, line 50: is it the Accessor's responsibility to *create* the store? that can be a function in our pipelines or elsewhere, that creates the store, and then loads it in a map. But for this module to know much about what to get from withiin a pusubstreams.Module.. goes too far imo.

pipeline/store/accessor.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (s *Accessor) Iter(cb func(name string, store *state.Store) error) error {
+	for name, store := range s.storeMap {
+		if err := cb(name, store); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
abourget, line 160: woah, we need a func with callback to loop through our map?
you know our initial StoreMap concept, could very well have revealed its internal `map` with a method that wouldn't be available to the `StoreGetter`

Initially, we wanted to cut the object in 3-4 parts.. but it ended all cobbled up in an object that doesn't have clear boundaries or responsibilities.

pipeline/store/accessor.go
```diff
@@ -0,0 +1,177 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to flush stores: %w", err)
+		}
+		out = append(out, blockRanges...)
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	return out, nil
+}
```
abourget, line 107: All this work! `Hydrate`, `Flush` .. which is business logic of the user.

I think we can give another run of analysis on that famous Accessor :)

pipeline/execout/cachev1/exec.go
```diff
@@ -0,0 +1,56 @@
...
+	return &ExecOutputCache{
+		engine: e,
+		clock:  clock,
+		cursor: cursor,
+		values: map[string][]byte{
+			blockType:                blkBytes,
+			"sf.substreams.v1.Clock": clockBytes,
+		},
+	}, nil
+}
```
abourget, line 29: This whole function's code is duplicated by the two implementations. It can be factored into a simple function, and called by the caller before calling NewExecOutput.

pipeline/forkhandler.go
```diff
@@ -22,19 +23,20 @@ func NewForkHandle() *ForkHandler {
...
 	respFunc func(resp *pbsubstreams.Response) error,
 ) error {
 	if moduleOutputs, found := f.reversibleOutputs[clock.Number]; found {
 		if err := returnModuleDataOutputs(clock, bstream.StepUndo, cursor, moduleOutputs, respFunc); err != nil {
 			return fmt.Errorf("calling return func when reverting outputs: %w", err)
 		}
 		for _, moduleOutput := range moduleOutputs {
-			if outputCache, ok := moduleOutputCache.OutputCaches[moduleOutput.Name]; ok {
-				outputCache.Delete(clock.Id)
+			outputsCache.UndoBlock(moduleOutput.Name, clock.Id)
```
abourget, line 35: Wouldn't it be useful for the `CacheEngine` to just be instructed to `UndoBlock()` with a block ID ?

What happens if we forget to loop on a certain module (we often check if `isOutput`).. That would ruin the internal stability of the CacheEngine. So it's his responsibility to loop through his modules and make sure he properly rolled out whatever he had. No?

pipeline/forkhandler.go
```diff
@@ -53,7 +55,4 @@ type DeltaGetter interface {
 }
 
 func reverseDeltas(storeMap map[string]*state.Store, name string, deltaGetter DeltaGetter) {
-	if store, found := storeMap[name]; found {
-		store.ApplyDeltaReverse(deltaGetter.GetDeltas())
-	}
 }
```
abourget, line 58: Is this function still useful?

pipeline/backprocess.go
```diff
@@ -3,118 +3,115 @@ package pipeline
...
+	p.reqCtx.StartSpan("back_processing")
+	successMessage := "completed"
+	defer func() {
+		if err != nil {
+			p.reqCtx.SetErr(err)
+		} else {
+			p.reqCtx.SetOk(successMessage)
+		}
+		p.reqCtx.EndSpan()
+	}()
```
abourget, line 28: I think this could be replaced with something much leaner:
```
defer p.reqCtx.StartSpan("back_processing").EndSpan()
defer p.reqCtx.SetReturn(&err)
```
and have `SetReturn` distinguish between a `nil` (and set `completed` by default, or I suggest `ok` as a simple affirmative, shorter, less data loaded in the tracer APIs).

One could even think of a return of a `SuccessMessage("my message")` that would be altered into a `nil` in the `*err` within `SetReturn`.

My suggestion though? Drop the `SetOk()` message altogether, unless we feel we _really_ need the distinguished success messages.

pipeline/execout/cachev1/engine.go
```diff
@@ -0,0 +1,143 @@
...
+		SaveBlockInterval: saveBlockInterval,
+		baseCacheStore:    baseCacheStore,
+		logger:            logger,
+	}
+	return e, nil
+}
+
+func (e *Engine) Prime(modules map[string]*pbsubstreams.Module) error {
+	for hash, module := range modules {
+		if err := e.registerCache(module, hash); err != nil {
```
abourget, line 46: This function only needs the `Name`, no need for the full module.

pipeline/execout/cachev1/engine.go
```diff
@@ -0,0 +1,143 @@
...
+		ctx:               ctx,
+		caches:            make(map[string]*OutputCacheState),
+		SaveBlockInterval: saveBlockInterval,
+		baseCacheStore:    baseCacheStore,
+		logger:            logger,
+	}
+	return e, nil
+}
+
+func (e *Engine) Prime(modules map[string]*pbsubstreams.Module) error {
```
abourget, line 44: This function knows too much about the `Module`, where it only needs a mapping between Name and Hash. Don't we have that with the ModuleHashes object?

That's the only concern of the CacheEngine.

I'd decouple from `pbsubstreams.Module`

pipeline/execout/cachev1/engine.go
```diff
@@ -0,0 +1,143 @@
...
+	logger            *zap.Logger
+}
+
+type OutputCacheState struct {
+	c           *OutputCache
+	initialized bool
+}
+
+func NewEngine(ctx context.Context, saveBlockInterval uint64, baseCacheStore dstore.Store, logger *zap.Logger) (execout.CacheEngine, error) {
+	e := &Engine{
```
abourget, line 34: Why not leverage the `reqCtx` here? We'd have the `ctx` and the `logger`.

pipeline/process_block.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	p.reqCtx.StartSpan("modules_executions")
+	for _, executor := range p.moduleExecutors {
+		if err = p.runExecutor(p.reqCtx.Context(), clock, executor, execOutput); err != nil {
+			return err
+		}
+	}
+	metrics.BlockEndProcess.Inc()
+	p.reqCtx.EndSpan()
```
abourget, line 115: wrap this execution in a function, it's possible we don't call `EndSpan()` if we return an err.. that would be an imbalance. Let's have that in a `defer` clause, and have the execution loop be in a function.

Small functions FTW

pipeline/process_block.go
```diff
@@ -0,0 +1,135 @@
...
+	if err = p.FlushStores(block.Num()); err != nil {
+		return fmt.Errorf("failed to flush stores: %w", err)
+	}
+
+	if isStopBlockReached(clock.Number, p.reqCtx.StopBlockNum()) {
+		p.reqCtx.logger.Debug("about to save cache output",
+			zap.Uint64("clock", clock.Number),
+			zap.Uint64("stop_block", p.reqCtx.StopBlockNum()),
+		)
+		if err = p.cachingEngine.Flush(p.reqCtx.Context()); err != nil {
```
abourget, line 102: could we haev `reqCtx` embed the `context.Context` ? So we could simply pass `reqCtx` where a `context.Context` is needed? It's an interface.

pipeline/context.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (r *RequestContext) StartSpan(spanName string) {
+	ctx, span := r.tracer.Start(r.ctx, spanName)
+	r.traces = append(r.traces, &Trace{
+		name: spanName,
+		ctx:  ctx,
+		span: span,
+	})
+}
```
abourget, line 43: If we return the `span` here, the caller can `defer reqctx.StartSpan("name").End()`

abourget: Also, I'd vote for a `currentSpan` and some `panic` if we don't unwind properly, or if we go too far.  This would mean structural errors in our code, and warrants a panic I think.

pipeline/execout/cachev1/engine.go
abourget: Wondering about `Prime`, what we prime? It's not prime number related either.

Isn't that simply an `Init`ialization?

pipeline/execout/interface.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+var NotFound = errors.New("inputs module value not found")
+
+type CacheEngine interface {
+	NewExecOutput(blockType string, block *bstream.Block, clock *pbsubstreams.Clock, cursor *bstream.Cursor) (ExecOutput, error)
+	Prime(modules map[string]*pbsubstreams.Module) error
+	UndoBlock(moduleName string, blockID string)
+	Flush(ctx context.Context) error
+	FlushAndUpdate(ctx context.Context, blockRef bstream.BlockRef) error
```
abourget, line 17: the caller ought not to know what the CacheEngine will do with this information, if it wants to Flush or not.. it's not the caller's concern.

If we have `UndoBlock` and that's the minimum thing we want to instruct the cache engine, then we should have something symmetric for the other block steps, like `NewBlock()` .. and let the engine decide whether to flush or not. It seems that that's what `FlushAndUpdate()` wants to be, doesn't it?

abourget: I'm unsure of the passing of `ctx context.Context` here also. The CacheEngine deals with its own context.. for example, you wouldn't want to cancel a flush because someone suddenly disconnects, right? The flush is not related to a particular request.

abourget: as for `Flush()` .. again it's not the concern of the caller what the action should be. The message being passed looks more like an `RangeEndReached()` .. which would come with a number that potentially is over a boundary, but has not been processed.. so it could inform the caching engine to flush what happened before.

So I imagine:

```
NewExecOutput()
Init(ModuleHashes)
NewBlock(blockRef)
UndoBlock(blockID string) // let UndoBlock do the looping
RangeEndReached(blockNum uint64)
```

abourget: There's also the issue that `RangeEndReached()` does not guarantee things to be irreversible, so shouldn't _necessarily_ trigger a Flush(), unless perhaps we also pass down the `Step`, and make a decision based on that.  See comments near `flushCache()`.

pipeline/execout/cachev1/engine.go
```diff
@@ -0,0 +1,143 @@
...
+		zap.String("module_name", cache.ModuleName),
+		zap.Uint64("start_block", cache.CurrentBlockRange.StartBlock),
+		zap.Uint64("end_block", cache.CurrentBlockRange.ExclusiveEndBlock),
+	)
+
+	if err := cache.save(ctx, filename); err != nil {
+		return fmt.Errorf("save: saving outpust or module kv %s: %w", cache.ModuleName, err)
+	}
+	return nil
+}
```
abourget, line 98: There's an important issue here we discussed but still isn't fixed. Perhaps it's documented somewhere else, but I'm dropping my notes here in case it isn't:

This function writes the output cache NO MATTER what state of the chain we're at. We should not save any files unless we're certain we've passed finality on all the blocked we added through `NewBlock()`.

We either need to be informed of finality through the steps handler of our ForkHandler (a custom one with all the steps).

Maybe this means therefore, that we should not have a `NewBlock()` but rather a `NewFinalBlock()` and make sure we never call this when we only have a `STEP_NEW`.  We don't want to fiddle with the cache when we're neaer real-time anyway, do we?!

pipeline/execout/cachev1/engine.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (e *Engine) registerCache(module *pbsubstreams.Module, hash string) error {
+	e.logger.Debug("registering modules", zap.String("module_name", module.Name))
+
+	if _, found := e.caches[module.Name]; found {
+		return nil
+	}
```
abourget, line 105: why not panic here? that's would be a programming error if we added the same module twice.

pipeline/execout/cachev1/engine.go
```diff
@@ -0,0 +1,143 @@
...
+	if !found {
+		return nil, false, fmt.Errorf("cache %q not found", moduleName)
+	}
+	if !cache.initialized {
+		// TODO: should context be passed upstream?
+		if _, err := cache.c.LoadAtBlock(e.ctx, clock.Number); err != nil {
+			return nil, false, fmt.Errorf("unable to load cache %q at block %d: %w", moduleName, clock.Number, err)
+		}
+		cache.initialized = true
+	}
```
abourget, line 130: we could decide to have the `initialized` flag set on the `Engine`, and loop through all of the uninitialized _caches_ and `LoadAtBlock` them in parallel.

The `ctx` for the caching engine ought to come from the top-most initialization of the process, shouldn't it? If we Ctrl+C, then perhaps we cancel the loads and saves, not based on the requests' contexts.

pipeline/execout/cachev1/engine.go
```diff
@@ -0,0 +1,143 @@
...
+
+	data, found := cache.c.Get(clock)
+	return data, found, nil
+}
+
+func (e *Engine) set(moduleName string, data []byte, clock *pbsubstreams.Clock, cursor string) error {
+	cache, found := e.caches[moduleName]
+	if !found {
+		return fmt.Errorf("cache %q not found", moduleName)
+	}
```
abourget, line 140: shouldn't that panic? that would be because of an inconsistent setup

pipeline/execout/inouts.go
```diff
@@ -0,0 +1,65 @@
...
+		values: map[string][]byte{
+			blockType:                blkBytes,
+			"sf.substreams.v1.Clock": clockBytes,
+		},
+	}, nil
+}
+
+type ExecOutputMap struct {
+	values map[string][]byte
+}
```
abourget, line 52: two structs in this file, perhaps `noopcache.go` for the NoOpCache?

pipeline/execout/cachev1/exec.go
```diff
@@ -0,0 +1,56 @@
...
+
+type ExecOutputCache struct {
+	engine *Engine
+	values map[string][]byte
+	clock  *pbsubstreams.Clock
+	cursor *bstream.Cursor
+}
```
abourget, line 36: thought this one could embed the `ExecOutputMap` from the `execout` package, and use its functions, and overload. If we add functionality to that ExecOutputMap, this one would benefit.

abourget: Also would show the relation between the two, one being the extension of the other

pipeline/pipeline.go
```diff
@@ -17,25 +16,21 @@ import (
...
 type Pipeline struct {
 	vmType    string // wasm/rust-v1, native
 	blockType string
 
-	requestedStartBlockNum uint64 // rename to: requestStartBlock, SET UPON receipt of the request
-	maxStoreSyncRangeSize  uint64
-	isSubrequest           bool
+	//requestedStartBlockNum uint64 // rename to: requestStartBlock, SET UPON receipt of the request
+	maxStoreSyncRangeSize uint64
+	isSubrequest          bool
```
abourget, line 33: why isn't `isSubrequest` part of the `requestContext` ? Isn't it context about the request?

service/validator.go
```diff
@@ -0,0 +1,49 @@
...
+	blockType string
+}
+
+func NewRequestValidator(blockType string) *RequestValidator {
+	return &RequestValidator{
+		blockType: blockType,
+	}
+}
+
+func (r *RequestValidator) Validate(request *pbsubstreams.Request) (*manifest.ModuleGraph, error) {
```
abourget, line 19: was the goal of this object not to have to store `blockType` on the `Service` object anymore?
Couldn't the `Validate()` method simply be passed the `blockType` from `s.blockType` ?

If this can be simply a function, that's easier than an object, especially if it's just to enclose a variable that is available at the moment of calling `Validate()`.

pipeline/context.go
abourget: With `currentSpan`, we could still add SetAttributes() to that span, anywhere in the scope of a `StartSpan()` before the `defer ..End()` is called, and it would add it to the current span.

service/service.go
```diff
@@ -120,12 +65,20 @@ func New(
...
+
 	zlog.Info("creating gprc client factory", zap.Reflect("config", substreamsClientConfig))
 	clientFactory := client.NewFactory(substreamsClientConfig)
 
-	s.workerPool, err = orchestrator.NewWorkerPool(parallelSubRequests, clientFactory)
+	workerFactoryGenerator := func() orchestrator.WorkerFactory {
+		return func(tracer ttrace.Tracer) orchestrator.Worker {
+			return orchestrator.NewRemoteWorker(clientFactory, tracer)
+		}
+	}
```
abourget, line 79: that seems a bit convoluted to pass down an instance of a tracer, isn't it? Couldn't that be passed as a variable somewhere in the stack?
The tracer instance is pretttty abstract and high up.. could be known any top sort of top-level objects.

Perhaps we're in need of an `Orchestrator` object? Something that would tie up all those loose ends in `backprocess.go` huh? :)

service/service.go
```diff
@@ -158,74 +123,43 @@ func (s *Service) Register(
...
-
-	sources := graph.GetSources()
-	for _, source := range sources {
-		if source != s.blockType && source != "sf.substreams.v1.Clock" {
-			err := fmt.Errorf(`input source %q not supported, only %q and "sf.substreams.v1.Clock" are valid`, source, s.blockType)
-			span.SetStatus(otelcode.Error, err.Error())
-			return err
-		}
+		return NewBasicErr(status.Error(codes.InvalidArgument, err.Error()), err)
 	}
```
abourget, line 155: have you guys checked what's in `derr` ? We have things to wrap/unwrap gRPC errors, etc..

pipeline/pipeline.go
```diff
@@ -132,312 +113,211 @@ func (p *Pipeline) isOutputModule(name string) bool {
...
+		)
 
-func (p *Pipeline) bumpStoreSaveBoundary() bool {
-	p.nextStoreSaveBoundary = p.computeNextStoreSaveBoundary(p.nextStoreSaveBoundary)
-	return p.request.StopBlockNum == p.nextStoreSaveBoundary
-}
-func (p *Pipeline) computeNextStoreSaveBoundary(fromBlock uint64) uint64 {
-	nextBoundary := fromBlock - fromBlock%p.storeSaveInterval + p.storeSaveInterval
-	if p.isSubrequest && p.request.StopBlockNum != 0 && p.request.StopBlockNum < nextBoundary {
-		return p.request.StopBlockNum
```
abourget, line 295: ouch I see those two methods removed, and some stuff in the Accessor that seems to do the same, although I remember needing to test these two methods with great care, to avoid a bunch of issues that I cannot recall.

What I see here is that some of those tests have simply been removed and not moved elsewhere. 
(searching `subreq, reqStart off bounds, current one off bound` doesn't show this test moved).

I probably didn't dig enough, but I'm a bit worried.

pipeline/store/accessor.go
```diff
@@ -0,0 +1,177 @@
...
+	return nil
+}
+
+func (r *Accessor) computeBoundaryBlock(atBlockNum uint64) uint64 {
+	boundary := atBlockNum - atBlockNum%r.SaveInterval + r.SaveInterval
+	if r.maxBoundaryBlockNum > 0 && r.maxBoundaryBlockNum < boundary {
+		return r.maxBoundaryBlockNum
+	}
+	return boundary
+}
```
abourget, line 173: I recall we needed the distinction between `initStoreSaveBoundary` and `bumpStoreSaveBoundary` and it was a subtle difference.

Now I only see `computeBoundaryBlock` and the tests that covered it were removed.

pipeline/store/accessor.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (r *Accessor) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	enc.AddInt("store_count", len(r.storeMap))
+	return nil
+}
+
+func (r *Accessor) computeBoundaryBlock(atBlockNum uint64) uint64 {
+	boundary := atBlockNum - atBlockNum%r.SaveInterval + r.SaveInterval
+	if r.maxBoundaryBlockNum > 0 && r.maxBoundaryBlockNum < boundary {
```
abourget, line 169: what about the `isSubrequest` clause here that was there previously? You sure it's not needed anymore?

pipeline/pipeline.go
```diff
@@ -132,312 +113,211 @@ func (p *Pipeline) isOutputModule(name string) bool {
...
 
-	for _, module := range p.modules {
-		if p.outputModuleMap[module.Name] {
-			p.moduleHashes.HashModule(p.request.Modules, module, p.graph)
-		}
+	if isOutput && p.reqCtx.StartBlockNum() < module.InitialBlock {
+		return fmt.Errorf("invalid request: start block %d smaller that request outputs for module: %q start block %d", p.reqCtx.StartBlockNum(), module.Name, module.InitialBlock)
 	}
+	p.moduleHashes.HashModule(p.reqCtx.Request().Modules, module, p.graph)
+	return nil
```
abourget, line 124: hmm, this is for a single module now?
Are we not resetting `moduleHashes` at each run then?
That's why we had it handle all the modules in one swift, so that we'd initialize the`NewModulesHashes()` once, and each call to `computeModuleHashes()` would be atomic, initialize and compute anew, in preparation for the time we'd need to recompute the hashes, we'd just call this atomic function and we'll all be set.

pipeline/pipeline.go
```diff
@@ -132,312 +113,211 @@ func (p *Pipeline) isOutputModule(name string) bool {
...
+	}
 
-		hash := p.moduleHashes.Get(module.Name)
-		_, err := p.moduleOutputCache.RegisterModule(module, hash, p.baseStateStore)
-		if err != nil {
-			return fmt.Errorf("registering output cache for module %q: %w", module.Name, err)
-		}
+	p.reqCtx.logger.Info("priming caching engine")
+	if err := p.cachingEngine.Prime(moduleHashes); err != nil {
+		return fmt.Errorf("failed to prime caching engine: %w", err)
```
abourget, line 164: unsure about the rational to pry open the modules hashing thing, which was all containerized in that `ModulesHashes` object, especially when I see that the caching engine would benefit from that exact abstraction.

pipeline/store/accessor.go
abourget: Or a `BoundedStore` ? which knows its boundaries?

There's a StoreReadWriter / StoreReader and StoreWriter in here.. that knows how to Load and Save itself from storage.

There's signals the StoreMap could dispatch to the Store or PartialStore depending.. so would relieve the Pipeline from knowing what will actually happen with the stores, when certain signals arrive (like BlockComplete or something).

Maybe we also think here that we only trigger on Final signals, because we don't want to save stores in the reversible segments. Is that propery handled?

Perhaps StoreMap has the same things as the ForkHandler, something like `Init`, `NewFinalBlock`, `UndoBlock` or whatnot? Same as the ExecOutput has? That would be nice, if every component knew what to do with the reorg signals.

pipeline/execout/cachev1/cache.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	sort.Slice(out, func(i, j int) bool {
+		return out[i].StartBlock < out[j].StartBlock
+	})
+
```
abourget, line 228: You can sort.Sort() directly the `block.Ranges` which has everything needed already, doesn't it? If need to be reversed, there's `sort.Reversed()` or something.

abourget: https://github.com/streamingfast/substreams/blob/bob/block/range.go#L119-L121

pipeline/execout/cachev1/cache.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+		if endBlock > biggestEndBlock {
+			biggestEndBlock = endBlock
+		}
+	}
+
+	exclusiveEndBlock = biggestEndBlock
+
+	return block.NewRange(prefixStartBlock, exclusiveEndBlock), true, nil
+}
```
abourget, line 306: I like these functions here. They abstract complexities away, hide them behind a meaningful name. And simplify the call site, all the while not introducing new objects with side effects mutating struct members.

Functions, functions functions! Let's do them more :)

pipeline/store/accessor.go
jubeless: I need more context here. What is the difference between 

`PartialStore`
`Store`
`StoreMap`

PR #53, opened 2022-10-14
Adding flag to specify output-dir for pack command

robinbernon:
Added an optional flag to specify output-dir for pack command. Also changed all occurrences of /tmp/ folder paths to use os.TempDir() instead to allow cli tool to work on windows OS.

### DIFF
--- a/cmd/substreams/pack.go
+++ b/cmd/substreams/pack.go
@@ -10,7 +10,6 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/streamingfast/cli"
 	"github.com/streamingfast/substreams/manifest"
-	"go.uber.org/zap"
 	"google.golang.org/protobuf/proto"
 )
 
@@ -25,31 +24,43 @@ var packCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(packCmd)
 	packCmd.Flags().StringP("output-dir", "o", ".", cli.FlagDescription(`
-		Optional flag to specify output directory to output generated .spkg file. If the received <package> argument is a local Substreams manifest file
-		(e.g. a local file ending with .yaml), the output folder will be made relative to it
+		Optional flag to specify output directory to output generated .spkg file. 
+		If the received argument starts with a "." it will be treated as a local path relative to the supplied manifest path.
 	`))
+}
+
+func is_relative_path(path string) bool {
+	return path[0:1] == "."
 }
 
 func runPack(cmd *cobra.Command, args []string) error {
 	outputDir := maybeGetString(cmd, "output-dir")
-
-	validOutputDirectorySpecified := true
-	fileInfo, err := os.Stat(outputDir)
-	if err != nil || !fileInfo.IsDir() {
-		fmt.Println("WARNING: Output directory specified is invalid - falling back to default path!\nOutputDir=", outputDir)
-		fmt.Println("")
-		validOutputDirectorySpecified = false
+	if outputDir == "." {
+		outputDir = ""
 	}
 
 	manifestPath := args[0]
 
-	manifestReader := manifest.NewReader(manifestPath)
+	if outputDir != "" {
+		if is_relative_path(outputDir) {
+			workingDirectory, err := os.Getwd()
+			if err != nil {
+				return fmt.Errorf("can't retrieve current directory information")
+			}
+			newOutputDir := filepath.Join(filepath.Dir(filepath.Join(workingDirectory, manifestPath)), outputDir)
+			fmt.Printf("Output directory specified: %s \nThis will be treated as a local path.\nFull folderpath: %s\n", outputDir, newOutputDir)
+			outputDir = newOutputDir
+		} else {
+			fmt.Printf("Output directory treated as an absolute path.\nFull folderpath: %s\n", outputDir)
+		}
 
-	if validOutputDirectorySpecified && manifestReader.IsLocalManifest() && !filepath.IsAbs(outputDir) {
-		newOutputDir := filepath.Join(filepath.Dir(manifestPath), outputDir)
-		zlog.Debug("manifest path is a local manifest, making output folder relative to it", zap.String("old", outputDir), zap.String("new", newOutputDir))
-		outputDir = newOutputDir
+		err := os.MkdirAll(outputDir, os.ModePerm)
+		if err != nil {
+			return fmt.Errorf("error creating output directory: %w", err)
+		}
 	}
+
+	manifestReader := manifest.NewReader(manifestPath)
 
 	pkg, err := manifestReader.Read()
 	if err != nil {
@@ -67,17 +78,10 @@ func runPack(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("marshalling package: %w", err)
 	}
 
-	if validOutputDirectorySpecified {
-		outputFile := filepath.Join(outputDir, defaultFilename)
-		if err := ioutil.WriteFile(outputFile, cnt, 0644); err != nil {
-			fmt.Println("")
-			return fmt.Errorf("writing %q: %w", outputFile, err)
-		}
-	} else {
-		if err := ioutil.WriteFile(defaultFilename, cnt, 0644); err != nil {
-			fmt.Println("")
-			return fmt.Errorf("writing %q: %w", defaultFilename, err)
-		}
+	outputFile := filepath.Join(outputDir, defaultFilename)
+	if err := ioutil.WriteFile(outputFile, cnt, 0644); err != nil {
+		fmt.Println("")
+		return fmt.Errorf("writing %w", err)
 	}
 
 	fmt.Printf(`To generate bindings for your code:
--- a/cmd/substreams/pack.go
+++ b/cmd/substreams/pack.go
@@ -29,10 +29,6 @@ func init() {
 	`))
 }
 
-func is_relative_path(path string) bool {
-	return path[0:1] == "."
-}
-
 func runPack(cmd *cobra.Command, args []string) error {
 	outputDir := maybeGetString(cmd, "output-dir")
 	if outputDir == "." {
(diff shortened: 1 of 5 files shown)


test/integration_test.go
```diff
@@ -109,7 +110,8 @@ func processRequest(t *testing.T, request *pbsubstreams.Request, moduleGraph *ma
 
 	req := pipeline.NewRequestContext(ctx, request, isSubRequest)
 
-	baseStoreStore, err := dstore.NewStore("file:///tmp/test.store", "", "none", true)
+	file := fmt.Sprintf("file://%s", filepath.Join(os.TempDir(), "test.store"))
+	baseStoreStore, err := dstore.NewStore(file, "", "none", true)
```
maoueh, line 114: Golang has a test facility to deal with temp dir in test, you can use it here (seems the original author didn't know about that either :D, I just told the team). Also the `file://` is superflous.
```suggestion
	baseStoreStore, err := dstore.NewStore(filepath.Join(t.TempDir(), "test.store"), "", "none", true)
```

test/integration_test.go
```diff
@@ -202,7 +204,8 @@ type AssertMapOutput struct {
 func runTest(t *testing.T, startBlock int64, exclusiveEndBlock uint64, moduleNames []string, newBlockGenerator NewTestBlockGenerator) (moduleOutputs []string) {
 	//_, _ = logging.ApplicationLogger("test", "test")
 
-	err := os.RemoveAll("/tmp/test.store")
+	filepath := filepath.Join(os.TempDir(), "test.store")
+	err := os.RemoveAll(filepath)
```
maoueh, line 208: And there is no need to remove the file anymore, `t.TempDir` is automatically cleaned up when the test (and its sub-tests) end

cmd/substreams/pack.go
```diff
@@ -20,12 +24,33 @@ var packCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(packCmd)
+	packCmd.Flags().StringP("output-dir", "o", "src/pb", cli.FlagDescription(`
+		Optional flag to specify output directory to output generated .spkg file. If the received <package> argument is a local Substreams manifest file
```
maoueh, line 28: Since a default value exists, all flags are technically optional, so let's change the wording a little bit
```suggestion
		Sspecify output directory where the output generated .spkg file. If the received <package> argument is a local Substreams manifest file
```

And let's remove the thing about relative path if we decide to remove that feature from `pack`.

cmd/substreams/pack.go
```diff
@@ -20,12 +24,33 @@ var packCmd = &cobra.Command{
...
+		(e.g. a local file ending with .yaml), the output folder will be made relative to it
+	`))
 }
 
 func runPack(cmd *cobra.Command, args []string) error {
+	outputDir := maybeGetString(cmd, "output-dir")
+
+	validOutputDirectorySpecified := true
+	fileInfo, err := os.Stat(outputDir)
+	if err != nil || !fileInfo.IsDir() {
```
maoueh, line 38: We should instead create the hierarchy of folders automatically with `os.MkdirAll` and not check if it exists and any I/O error will be caught right now in `os.MkdirAll` or later when writing the file.

maoueh (changes requested):
Add also a CHANGELOG notice about the new flag

cmd/substreams/pack.go
robinbernon: Well yes, I agree they usually are. However the protogen cmd has two "mandatory flags" that need to be specified so I thought it best to distinguish between the two.

cmd/substreams/pack.go
```diff
@@ -20,12 +24,33 @@ var packCmd = &cobra.Command{
...
+		fmt.Println("WARNING: Output directory specified is invalid - falling back to default path!\nOutputDir=", outputDir)
+		fmt.Println("")
+		validOutputDirectorySpecified = false
+	}
+
 	manifestPath := args[0]
 
 	manifestReader := manifest.NewReader(manifestPath)
+
+	if validOutputDirectorySpecified && manifestReader.IsLocalManifest() && !filepath.IsAbs(outputDir) {
```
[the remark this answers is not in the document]
robinbernon: Definitely want to be able to specify relative directory paths as it's a pain to have to work with absolute paths cross-platform when calling it from places like makefiles. Was new to golang do didn't realise what the right command was but essentially we'd need to check that for a folder specified, is it specified from the root folder of your operating system or not. If not then we will treat it like a relative path. I guess I wouldn't really want to decide whether it's a relative path or not depending on whether the manifest was local anyway.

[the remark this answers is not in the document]
robinbernon: I guess a simple way to do this would be just to check whether the first character of the supplied path is '.' or not.. what would you think about doing that?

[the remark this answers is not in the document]
maoueh: `filepath.IsAbs(<path>)` can be used to determine if a path is relative already or absolute. Relative path handling will of course continue to work, the question is more abou what is the best default behavior for `substreams pack`:

- Relative to `manifest.yaml` file
- Relative to working directory of the process (so usually the path in your terminal where the command is invoked).

I think for `substreams pack`, the best default is relative to command execution, scripting wise it's the easiest. It's also the easiest to reason about when passing the flag, specially for project with multiple substreams:

```
substreams pack -o build/spkgs substreams1/manifest.yaml
```

Is it `./build/spkgs` or `substreams1/build/spkgs/manifest.yaml`? I had this debate internally even for `substreams protogen` but in this particular case, it made sense to make it relative to `manifest.yaml` (when `IsLocalmanifest`) because the generated protobuf files are meant to be used in this particular substreams project.

Not so much the case with `pack`.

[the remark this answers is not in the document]
robinbernon: I've made what I think is a good change if you want to have a look? Think this should be the clearest way to deal with this.

cmd/substreams/pack.go
```diff
@@ -20,12 +24,33 @@ var packCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(packCmd)
+	packCmd.Flags().StringP("output-dir", "o", "src/pb", cli.FlagDescription(`
```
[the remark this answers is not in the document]
robinbernon: DOne

cmd/substreams/pack.go
```diff
@@ -42,9 +67,17 @@ func runPack(cmd *cobra.Command, args []string) error {
...
+		outputFile := filepath.Join(outputDir, defaultFilename)
+		if err := ioutil.WriteFile(outputFile, cnt, 0644); err != nil {
+			fmt.Println("")
+			return fmt.Errorf("writing %q: %w", defaultFilename, err)
+		}
+	} else {
+		if err := ioutil.WriteFile(defaultFilename, cnt, 0644); err != nil {
+			fmt.Println("")
+			return fmt.Errorf("writing %q: %w", defaultFilename, err)
+		}
```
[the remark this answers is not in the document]
robinbernon: Done

cmd/substreams/pack.go
```diff
@@ -42,9 +67,17 @@ func runPack(cmd *cobra.Command, args []string) error {
...
 	}
 
-	if err := ioutil.WriteFile(defaultFilename, cnt, 0644); err != nil {
-		fmt.Println("")
-		return fmt.Errorf("writing %q: %w", defaultFilename, err)
+	if validOutputDirectorySpecified {
+		outputFile := filepath.Join(outputDir, defaultFilename)
+		if err := ioutil.WriteFile(outputFile, cnt, 0644); err != nil {
+			fmt.Println("")
+			return fmt.Errorf("writing %q: %w", defaultFilename, err)
```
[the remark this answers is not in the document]
robinbernon: Done

test/integration_test.go
robinbernon: Done

test/integration_test.go
robinbernon: Done

cmd/substreams/pack.go
robinbernon: 👍

cmd/substreams/pack.go
robinbernon: 👍

cmd/substreams/pack.go
```diff
@@ -20,12 +23,45 @@ var packCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(packCmd)
+	packCmd.Flags().StringP("output-dir", "o", ".", cli.FlagDescription(`
+		Optional flag to specify output directory to output generated .spkg file. 
+		If the received argument starts with a "." it will be treated as a local path relative to the supplied manifest path.
+	`))
+}
+
+func is_relative_path(path string) bool {
```
maoueh, line 32: Golang uses `camelCase` for private method and `PascalCase` for public methods, so should have been `isRelativePath`.

Second, this method is not needed, call site should be replace with `!filepath.IsAbs(path)`

cmd/substreams/pack.go
```diff
@@ -20,12 +23,45 @@ var packCmd = &cobra.Command{
...
+	return path[0:1] == "."
 }
 
 func runPack(cmd *cobra.Command, args []string) error {
+	outputDir := maybeGetString(cmd, "output-dir")
+	if outputDir == "." {
+		outputDir = ""
+	}
```
maoueh, line 40: Not needed anymore, reality is that "." and "" are equivalent when making relative path absolute base on working directory see other bigger comment.
```suggestion
```

cmd/substreams/pack.go
```diff
@@ -20,12 +23,45 @@ var packCmd = &cobra.Command{
...
+			outputDir = newOutputDir
+		} else {
+			fmt.Printf("Output directory treated as an absolute path.\nFull folderpath: %s\n", outputDir)
+		}
+
+		err := os.MkdirAll(outputDir, os.ModePerm)
+		if err != nil {
+			return fmt.Errorf("error creating output directory: %w", err)
+		}
+	}
```
maoueh, line 61: All this is not needed, `os.MkdirAll` and `ioutil.WriteFile` (and any other I/O path related functions) automatically make relative path absolute against the process working directory, so it's not needed at all to make the path absolute.
```suggestion
        if err := os.MkdirAll(outputDir, os.ModePerm); err != nil {
	        return fmt.Errorf("creating output directories: %w", err)
        }
```

cmd/substreams/pack.go
```diff
@@ -42,9 +78,10 @@ func runPack(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("marshalling package: %w", err)
 	}
 
-	if err := ioutil.WriteFile(defaultFilename, cnt, 0644); err != nil {
+	outputFile := filepath.Join(outputDir, defaultFilename)
+	if err := ioutil.WriteFile(outputFile, cnt, 0644); err != nil {
 		fmt.Println("")
-		return fmt.Errorf("writing %q: %w", defaultFilename, err)
+		return fmt.Errorf("writing %w", err)
```
maoueh, line 84: The `%w` wraps another error creating a chain of errors, convention in Go is to always separate contextual message `writing` here and the wrapped error `err` with a `:` character (and `:` should only be used once in the string and for this purposes). When the full error reaches the user/developer, it easy to spot the "error chain":

```
unable to pack: writing file: <some I/O error with the path in it>
```
```suggestion
		return fmt.Errorf("writing file: %w", err)
```

maoueh (changes requested):

cmd/substreams/pack.go
robinbernon: Have changed, thanks for the suggestion

cmd/substreams/pack.go
robinbernon: Have now removed

robinbernon:
> Add also a CHANGELOG notice about the new flag

Have added:)

maoueh:
I've merged your PR and built a different version of it. I've turned `output-dir` into `output-file` and added possibility to use two "template" directives to define the value. The default value of the flag is now `{manifestDir}/{spkgDefaultName}` which means old behavior of `substreams pack <manifest>` is preserved. And now scripting also works properly and user can decide where they want to emit the file and it's easy to switch from path relative to working directory (by not using `{manifestDir}` and also make the path relative to manifest dir with something like `{manifestDir}/build/mycustomname.spkg`.

PR #71, opened 2022-12-15
Updates for Various small tweaks for Documentation#64

seanmooretechwriter:
Addressing updates for the task:

Throughout https://substreams.streamingfast.io/developer-guide/, there is some old module name reference mainly block_to_transfers and nft_state, they need to be changed because they are not aligned with substreams-template. All occurrences should be renamed Update also https://substreams.streamingfast.io/developer-guide/creating-your-manifest#module-definitions so that modules have either map_ and store_ prefixes, we have settled more or less for this convention. We also need to have this loose "convention" documented at various places in the documentation maybe, maybe as a "note" element. Not sure exactly where, probably where first introduction to creating a new module is made and maybe in "reference" page.

### DIFF
--- a/docs/developer-guide/creating-your-manifest.md
+++ b/docs/developer-guide/creating-your-manifest.md
@@ -97,7 +97,7 @@ The manifest defines a list of [modules](../concepts/modules.md) used in the Sub
 The modules are Rust functions containing the business logic for the implementation.
 
 {% hint style="info" %}
-**Note**: The manifest in the Substreams Template example lists two modules: `map_transfers` and `store_transfers.` The naming convention prescribed by StreamingFast for Substreams modules is to prefix the name with either `map_` or `store_` depending on the module type.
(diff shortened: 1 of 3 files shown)


docs/developer-guide/creating-your-manifest.md
```diff
@@ -96,29 +97,29 @@ The manifest defines a list of [modules](../concepts/modules.md) used in the Sub
 The modules are Rust functions containing the business logic for the implementation.
 
 {% hint style="info" %}
-**Note**: The manifest in the Substreams Template example lists two modules: `block_to_transfers` and `nft_state.`
+**Note**: The manifest in the Substreams Template example lists two modules: `map_transfers` and `store_transfers.` The naming convention prescribed by StreamingFast for Substreams modules is to prefix the name with either `map_` or `store_` depending on the module type.
```
maoueh, line 100: > The naming convention prescribed by StreamingFast

Let's try to not mention `StreamingFast` when not required, it reads kind of weird to me.

```suggestion
**Note**: The manifest in the Substreams Template example lists two modules: `map_transfers` and `store_transfers.` The naming convention for Substreams modules is to prefix the name with either `map_` or `store_` depending on the module type.
```

maoueh (changes requested):

PR #74, opened 2022-12-15
Store size limit to 1GiB

sduchesneau:
* store total size is now limited to 1GiB.
* TotalSizeBytes is calculated on load (during the unmarshal) and modified on every kv operation to avoid recalculation

### DIFF
--- a/storage/store/merge.go
+++ b/storage/store/merge.go
@@ -12,6 +12,21 @@ import (
 	pbsubstreams "github.com/streamingfast/substreams/pb/sf/substreams/v1"
 )
 
+func (b *baseStore) setKV(k string, v []byte) {
+	if prev, ok := b.kv[k]; ok {
+		b.totalSizeBytes -= uint64(len(prev))
+	} else {
+		b.totalSizeBytes += uint64(len(k))
+	}
+	b.totalSizeBytes += uint64(len(v))
+	b.kv[k] = v
+}
+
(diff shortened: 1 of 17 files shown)


maoueh:
@sduchesneau Please add changelog entry for this change

storage/store/merge.go
```diff
@@ -33,28 +48,28 @@ func (b *baseStore) Merge(kvPartialStore *PartialKV) error {
 	switch b.updatePolicy {
 	case pbsubstreams.Module_KindStore_UPDATE_POLICY_SET:
 		for k, v := range kvPartialStore.kv {
-			b.kv[k] = v
+			b.setKV(k, v)
```
abourget, line 51: `updateKV` to clarify intent, different from `setNewKV` ? Someone could easily mix `setKV` and `setNewKV` not knowing the difference.

storage/store/merge.go
```diff
@@ -12,6 +12,21 @@ import (
 	pbsubstreams "github.com/streamingfast/substreams/pb/sf/substreams/v1"
 )
 
+func (b *baseStore) setKV(k string, v []byte) {
+	if prev, ok := b.kv[k]; ok {
```
abourget, line 16: This does add a _second_ lookup though; 
The caller of `setKV` often did it itself.

We can live with it for now I think.

abourget:
Looked good in general otherwise ;)

PR #102, opened 2022-12-22
dev mode output all modules (filter in cli)

### LINKED ISSUE #82: Activate all output of all modules in development mode
In this task:
- Ensure that when `development` mode is active, every module output is returned back rom the server
- In the Substreams CLI, by default, show only the output of the root module specified by the user
- In the Substreams CLI, accept a new parameter `--show-only-module-output` (name can be something else, it's the best I thought of so happy if it can be improved). This new parameter when set will change which module's output is printed, instead of being only the root module like by default, only those listed here will be specified. Should accept multiple modules separated by comma.

Edit the `CHANGELOG` to ensure this is properly documented, I don't want to see a single line, I want to see great details of how everything works and standard debug use cases.

sduchesneau:
fixes #82 
* server sends all module outputs and logs when in dev mode
* cli filters based on the output module or debug-modules-output
* add flag --debug-modules-output
* change (broken) flag --debug-modules-initial-snapshot to a list of modules

### DIFF
--- a/cmd/substreams/run.go
+++ b/cmd/substreams/run.go
@@ -58,9 +58,6 @@ func runRun(cmd *cobra.Command, args []string) error {
 	}
 
 	debugModulesInitialSnapshot := mustGetStringSlice(cmd, "debug-modules-initial-snapshot")
-	if debugModulesInitialSnapshot != nil && productionMode {
-		return fmt.Errorf("cannot set 'debug-modules-initial-snapshot' in 'production-mode'")
-	}
 
 	graph, err := manifest.NewModuleGraph(pkg.Modules.Modules)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


cmd/substreams/run.go
```diff
@@ -50,13 +51,24 @@ func runRun(cmd *cobra.Command, args []string) error {
...
+	productionMode := mustGetBool(cmd, "production-mode")
+	debugModulesOutput := mustGetStringSlice(cmd, "debug-modules-output")
+	if debugModulesOutput != nil && productionMode {
+		return fmt.Errorf("cannot set 'debug-modules-output' in 'production-mode'")
+	}
+
+	debugModulesInitialSnapshot := mustGetStringSlice(cmd, "debug-modules-initial-snapshot")
+	if debugModulesInitialSnapshot != nil && productionMode {
+		return fmt.Errorf("cannot set 'debug-modules-initial-snapshot' in 'production-mode'")
+	}
```
jubeless, line 63: I would not do this validation here, I would move this into the request validation, So that everyone else can benefit from it. That is what colin was working on. So you create the request naively and on line 111 it will fail out there.

Issue #135, opened 2023-01-10
Auto-infer `substreams.yaml` when using Substreams CLI


maoueh:
For each command that today's accept `substreams.yaml` at the end, the parameter should become optional. When not present, we would look in the current working directory for a file named `substreams.yaml` and use it directly if found. If nothing is provided and no file is found, error wrong usage of command.

If something is provided and it's a file that exists, use it right away. If something is provided, and that this parameter is an existing folder and a file `substreams.yaml` exists in this folder, use it.

The `substreams run` might prove a little bit trickier because the manifest is first and the second the module name. I suggest we deal with this using variable argument count `1-2` and check conditions

- [x] Update Cobra `short` usage to properly describe the command
- [x] Align Cobra `Args` to be strictly aligned with command accepted args
- [x] Ensure all commands are updated
- [x] Ensure you have tested all error paths and happy path, ensure you test with an `.spkg` also for the command.
- [x] Update the doc at https://substreams.streamingfast.io  accordingly, not sure where yet, will need to find the right spot(s)!

Relates to https://github.com/streamingfast/substreams/issues/122

PR #146, opened 2023-01-16
optional-yaml-param

emiliocramer:


tools/infer_manifest_test.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			var file *os.File
+			var dir string
+			var err error
+
+			if test.createsDir {
+				dir, _ = os.MkdirTemp("", test.inputString)
```
maoueh, line 35: The `t *testing.T` has an helper `t.TempDir()` that you can use, it will be automatically cleaned up when done with it.

tools/infer_manifest_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+				if test.createsFile {
+					file, _ = os.CreateTemp(dir, "substreams.yaml")
+				}
+
+				if file != nil {
+					_, err = os.ReadFile(file.Name())
```
maoueh, line 42: The test should create the "environment", not sure why a file is read in the test.

tools/infer_manifest_test.go
maoueh: So, you need to call `ResolveManifestFile` somewhere, it's the purpose of the test, so general flow is:

- Prepare test environment
- Call your function with test case input(s)
- Validate output so it matches the expected values
- Clean up if needed

I see no call to `ResolveManifestFile` here, so not sure what we are testing :)

tools/cmd.go
```diff
@@ -67,3 +70,30 @@ func mustGetDuration(cmd *cobra.Command, flagName string) time.Duration {
 	}
 	return val
 }
+
+func ResolveManifestFile(input string) (manifestName string, err error) {
+	if input == "" {
+		if _, err := os.Stat("substreams.yaml"); err == nil {
+			return "substreams.yaml", nil
+		} else if errors.Is(err, os.ErrNotExist) {
```
maoueh, line 78: What is `err != nil && !errors.Is(err, os.ErrNotExist)`?

tools/cmd.go
```diff
@@ -67,3 +70,30 @@ func mustGetDuration(cmd *cobra.Command, flagName string) time.Duration {
...
+			return "", fmt.Errorf("finding manifest file: %w", err)
+		}
+	} else {
+
+		// Check valid file first
+		if _, err := os.Stat(input); err == nil {
+			_, err := os.Open(input)
+			if err == nil {
+				return input, nil
+			}
```
maoueh, line 88: You should first investigate the method you are using, for example `os.Stat` to see if something could be useful. After that, I also suggest that you perform some Google search for the specific problem your are tackling.

Opening a path to check what kind it is is clearly not correct.

maoueh:
Closes #135

cmd/substreams/codegen.go
```diff
@@ -12,18 +14,30 @@ import (
...
 
 var codegenCmd = &cobra.Command{
-	Use:   "codegen <package>",
+	Use:   "codegen [<manifest>]",
 	Short: "Generate a Rust trait and boilerplate code from your 'substreams.yaml' for nicer development",
-	RunE:  runCodeGen,
-	Args:  cobra.ExactArgs(1),
+	Long: cli.Dedent(`
+		Generate a Rust trait and boilerplate code from your 'substreams.yaml' for nicer development. 
+		The manifest is optional as it will try to find one in your pwd if nothing entered.
```
maoueh, line 21:
```suggestion
		The manifest is optional as it will try to find one a file named 'substreams.yaml' in current working directory if nothing entered.
```

cmd/substreams/codegen.go
```diff
@@ -12,18 +14,30 @@ import (
...
+		You may enter a dir that contains a 'substreams.yaml' file in place of <manifest_file>
```
maoueh, line 22:
```suggestion
		You may enter a directory that contains a 'substreams.yaml' file in place of '<manifest_file>'.
```

cmd/substreams/graph.go
```diff
@@ -2,16 +2,22 @@ package main
...
-	Use:          "graph <manifest_file>",
-	Short:        "GenerateProto mermaid-js graph document",
-	RunE:         runManifestGraph,
-	Args:         cobra.ExactArgs(1),
+	Use:   "graph [<manifest_file>]",
+	Short: "GenerateProto mermaid-js graph document",
+	RunE:  runManifestGraph,
+	Long: cli.Dedent(`
+		GenerateProto mermaid-js graph document. The manifest is optional as it will try to find one in your pwd if nothing entered.
+		You may enter a dir that contains a 'substreams.yaml' file in place of <manifest_file>
```
maoueh, line 18: We should probably have the same description for the "optional" description for each command. The current `codegen` seems better, I would just merge there the two sentences in a single "paragraph".

```
The manifest is optional as it will try to find one a file named 'substreams.yaml' in current working directory if nothing entered. You may enter a directory that contains a 'substreams.yaml' file in place of '<manifest_file>'.
```

cmd/substreams/run.go
```diff
@@ -32,19 +34,31 @@ func init() {
...
+	manifestPathRaw := ""
+
+	if len(args) == 2 {
+		manifestPathRaw = args[0]
+		args = args[1:]
+	}
+	manifestPath, err := tools.ResolveManifestFile(manifestPathRaw)
+	if err != nil {
+		return fmt.Errorf("resolving manifest: %w", err)
+	}
```
maoueh, line 61: What is the current behavior on the `run` command if the user makes the mistake of forgetting to put the `<module>` so `substreams run substreams.yaml -e mainnet....`. I would like that we improve a bit this experience event if maybe rare.

We can be a little more clever, if there is a single argument and it contains a '.' or it's a valid directory, we know we are missing the output module arg and we can improve the improve the error message.

We have helper `cli.DirectoryExists(path string)` that you can quickly use to determine if the input is a directory (returns false on 'error' which is good enough for our use case)

docs/reference-and-specs/command-line-interface.md
```diff
@@ -11,11 +11,13 @@ The `substreams` command line interface (CLI) is the primary user interface and
 The `substreams` CLI exposes many commands to developers enabling a range of features.&#x20;
 
 {% hint style="info" %}
-**Note**: When a package is specified you can use:&#x20;
+**Note**: When a package is specified, it is optional. If you do use it, you can use:&#x20;
 
 * Local `substreams.yaml` configuration files
 * Local `.spkg` package files
 * Remote `.spkg` package URLs
+* Local directory containing a `substreams.yaml` file<br />
```
maoueh, line 19: The `<br />` is not needed I'm pretty sure.

docs/reference-and-specs/command-line-interface.md
```diff
@@ -11,11 +11,13 @@ The `substreams` command line interface (CLI) is the primary user interface and
...
+  If you choose to not use it, make sure that you are in a directory that contains a substreams.yaml file. Otherwise, you will get a usage error back.
```
maoueh, line 20: > If you choose to not use it

The sentence is kind of attached to the bullet list item because it's indented. Remove indentation and add a blank line between line 19 & 20.

tools/cmd.go
```diff
@@ -67,3 +70,33 @@ func mustGetDuration(cmd *cobra.Command, flagName string) time.Duration {
 	}
 	return val
 }
+
+func ResolveManifestFile(input string) (manifestName string, err error) {
+	if input == "" {
+		_, err := os.Stat("substreams.yaml")
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				return "", fmt.Errorf("no manifest entered in dir w/o a manifest")
```
maoueh, line 79: Don't use text shortcuts like `w/o` for error messages and user facing messages as well as `dir`. There is no good reason to shorten words for user facing elements.

tools/cmd.go
```diff
@@ -67,3 +70,33 @@ func mustGetDuration(cmd *cobra.Command, flagName string) time.Duration {
...
+			}
+			return "", fmt.Errorf("finding manifest: %w", err)
+		}
+
+		return "substreams.yaml", nil
+	}
+
+	inputInfo, err := os.Stat(input)
+	if err != nil {
+		return "", fmt.Errorf("reading input: %w", err)
```
maoueh, line 89: > reading input

Feels like we are trying to read some content, maybe `read input file info: %w`.

tools/cmd.go
```diff
@@ -67,3 +70,33 @@ func mustGetDuration(cmd *cobra.Command, flagName string) time.Duration {
...
+	}
+
+	// If input is dir
```
maoueh, line 92: I think the conditional `if inputInfo.IsDir() {` is obvious already.

tools/cmd.go
```diff
@@ -67,3 +70,33 @@ func mustGetDuration(cmd *cobra.Command, flagName string) time.Duration {
...
+	if inputInfo.IsDir() {
+		potentialManifest := filepath.Join(inputInfo.Name(), "substreams.yaml")
+		_, err := os.Stat(potentialManifest)
+		if err != nil {
+			return "", fmt.Errorf("finding manifest in dir: %w", err)
```
maoueh, line 97:
```suggestion
			return "", fmt.Errorf("finding manifest in directory: %w", err)
```

tools/cmd.go
```diff
@@ -67,3 +70,33 @@ func mustGetDuration(cmd *cobra.Command, flagName string) time.Duration {
...
+		}
+		return "substreams.yaml", nil
```
maoueh, line 99: That's seems quite wrong ... If the input was a directory, so maybe `"/tmp/some/path"` we return `substreams.yaml` which points to current working directory, have you tested all cases properly?

tools/resolve_manifest_file_test.go
```diff
@@ -0,0 +1,109 @@
...
+		},
+		{
+			"input provided, invalid manifest file",
+			args{"substreams-custom.yaml", "", []string{}},
+			"",
+			errorEqual("reading input: stat substreams-custom.yaml: no such file or directory"),
+		},
+		{
+			"input provided, valid dir",
+			args{"manifests-dir", "manifests-dir", []string{"substreams.yaml"}},
```
maoueh, line 58: `"manifests-dir", "manifests-dir"` makes it hard to reason about args, `"manifests-dir", "sub-dir"` helps a bit

maoueh (changes requested):

docs/reference-and-specs/command-line-interface.md
emiliocramer: thats what I thought too but in the ide preview it did, not sure. I'll get rid of it

tools/cmd.go
emiliocramer: it didnt seem to be when i was testing the function. os.Stat was giving no error whether file or directory, which i also found weird so I added it

tools/resolve_manifest_file_test.go
emiliocramer: thats not the intended purpose, but highlights the problem that my variable name is not as accurate as it should be. Fixed

tools/cmd.go
maoueh: I'm talking about the comment here, to remove it since `if inputInfo.IsDir() {` says the same as `// If input is dir`, sorry should have been clearer

tools/resolve_manifest_file_test.go
maoueh: Ok good

tools/cmd.go
emiliocramer: all good ill get rid of it

tools/cmd.go
emiliocramer: yea you're right.

tools/cmd.go
```diff
@@ -67,3 +70,32 @@ func mustGetDuration(cmd *cobra.Command, flagName string) time.Duration {
...
+
+	if inputInfo.IsDir() {
+		potentialManifest := filepath.Join(inputInfo.Name(), "substreams.yaml")
+		_, err := os.Stat(potentialManifest)
+		if err != nil {
+			return "", fmt.Errorf("finding manifest in directory: %w", err)
+		}
+		return fmt.Sprintf("%s/substreams.yaml", input), nil
+	}
+	return inputInfo.Name(), nil
```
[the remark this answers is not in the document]
emiliocramer: ig it would make sense at the very end to just print 'input'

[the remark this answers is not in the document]
maoueh:
```
Name() string       // base name of the file
```

So yeah, it was wrong. I think that when writing the unit test code, we should have make a final validation to read the returned file to ensure the read works, that would have catch those mistake.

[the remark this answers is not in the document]
emiliocramer: changed to now just output input. if they enter `/tmp/project/file.yaml` and its checked as valid with `os.Stat` and its not a dir, the full path will be returned

PR #149, opened 2023-01-18
remove distinction between execOutput and storeSnapshots SaveInterval

sduchesneau:
they must be aligned into the new CacheSaveInterval. This is to prevent a tier2 trying to start at block 300 if the stores only exist at 200 or 400. Too much complexity added in generating the outputs on tier2 in production mode.

### DIFF
--- a/orchestrator/multisquasher.go
+++ b/orchestrator/multisquasher.go
@@ -56,7 +56,7 @@ func NewMultiSquasher(
 				return nil, fmt.Errorf("store %q not found", storeModuleName)
 			}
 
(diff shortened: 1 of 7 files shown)


sduchesneau:
It will currently result in 
```Error: rpc error: code = Internal desc = error building pipeline: failed setup request: parallel processing run: scheduler run: process job result for target "graph_out": worker ended in error: receiving stream resp: rpc error: code = Internal desc = error building pipeline: failed to load stores: load full store: load full store store_tokens_whitelist_pools at 0012374870-0012369621.kv: not found
```

with those settings:
```
  substreams-output-cache-save-interval: 10
  substreams-stores-save-interval: 20
```

colindickson (approved):

Issue #38, opened 2022-08-12
Params to modules / dynamic inputs


jubeless:
Currently Substreams do not allow  the abiliy to add dynamic inputs when running a substreams. To accomodate this feature we suggest the following modification to the substreams manifest

```yaml
imports:
  ethereum: substreams-ethereum-v1.0.0.spkg

modules:
  - name: mymod
    doc: """
      The params are URL-encoded key/value and deal with it. Two are supported: contract & account. They mean whatever
    """
    inputs:
      - params: string
      - source: sf.ethereum.type.v1.Block

params:
  "ethereum:othermod": override-params-optionally-copy-the-original-value-dude
  mymod: params
```

and the CLI could:

```bash
-p ethereum:othermod=original-value-plus-yours -p mymod=0x123123123
```

To be parameterizable, the `params` input needs to be present, and be the first input.

In the Rust module, it's a new input, decoded as a simple String (we need to tweak the `substreams-rs` macro to adapt to that).

The `wasm` runtime needs to pass the `params` value, like another input.

For now, we only support `string` as a `params`, and encourage people to use URL-encoding for more complex parameterization.  However, a simple contract address for instance, is totally fine.

---

The modifications to the `[substreams/v1/modules.proto](https://github.com/streamingfast/substreams/blob/develop/proto/sf/substreams/v1/modules.proto)` file woud look like this:

```protobuf

  message Input {
    oneof input {
      Source source = 1;
      Map map = 2;
      Store store = 3;
      Params params = 4;  # This added
    }

    message Params {  # This added
      string value = 1;
    }
```

This way, there is a way for default values to be lodged when publishing a module. It's already at the right place for the runtime to consume. And it can be overridden by client code as such:
 
Python code to tweak params [would live here](https://github.com/streamingfast/substreams-playground/blob/master/consumers/python/main.py#L34) and look like:

```python
[mod for mod in pkg.modules.modules if mod.name == "store_pools"][0].inputs[0].params.value = "myvalue"
```

abourget:
The other option was to use `params` as a string, the manifest be:
```
modules:
- name: mymod
  inputs:
  - params: string
  - eth.Block
  params: "default string"
```

with support on the CLI for something like:

```bash
-p mymod=hello=world&other=var \
-p othermod=0x123123123
```

abourget:
Use case with Antelope:
* goal is to add a field in the final output module that's going to a sink, to add the chain_id to certain objects.
[...]

PR #167, opened 2023-02-03
Input parameters

abourget:
Addresses: https://github.com/streamingfast/substreams/issues/38
by @colindickson


cmd/substreams/run.go
```diff
@@ -29,6 +29,7 @@ func init() {
 	runCmd.Flags().StringSlice("debug-modules-initial-snapshot", nil, "List of 'store' modules from which to print the initial data snapshot (Unavailable in Production Mode")
 	runCmd.Flags().StringSlice("debug-modules-output", nil, "List of extra modules from which to print outputs, deltas and logs (Unavailable in Production Mode)")
 	runCmd.Flags().Bool("production-mode", false, "Enable Production Mode, with high-speed parallel processing")
+	runCmd.Flags().StringSliceP("params", "x", nil, "Set a parameters for supported modules. Can be used multiple times. Example: -x module1=valA -x module2=valX&valY")
```
abourget: Here I propose we rather remove `-p` and `-k` and keep the long-form args for those flags, and we use `-p` here, for what needs to be more intuitive. The other ones are mostly for internal dev by the core team, in wonky environments.

cmd/substreams/run.go
```diff
@@ -124,6 +127,7 @@ func runRun(cmd *cobra.Command, args []string) error {
 		OutputModules:                       []string{outputModule}, //added for backwards compatibility, will be removed
 		ProductionMode:                      productionMode,
 		DebugInitialStoreSnapshotForModules: debugModulesInitialSnapshot,
+		Params:                              params,
```
abourget: There wouldn't be `Params` on the `pbsubstreams.Request`. Its place is on `Module_Input_Params` (as it is already).

manifest/manifest.go
```diff
@@ -138,7 +149,11 @@ func (i *Input) parse() error {
 		i.Name = fmt.Sprintf("source:%s", i.Source)
 		return nil
 	}
-	return fmt.Errorf("input has an unknown type. Expect one, and only one of 'map', 'store' or 'source'")
+	if i.IsParams() {
+		i.Name = fmt.Sprintf("params:%s", i.Params)
```
abourget: Shouldn't we validate here that only `string` is acceptable as a parameter? Unless it's done elsewhere, this should validate it.

manifest/manifest.go
```diff
@@ -281,6 +295,21 @@ func (m *Module) setInputsToProto(pbModule *pbsubstreams.Module) error {
...
 		}
+		if input.Params != "" {
+			if i != 0 {
+				return fmt.Errorf("input.params must be the first input")
+			}
+
+			pbInput := &pbsubstreams.Module_Input{
+				Input: &pbsubstreams.Module_Input_Params_{
+					Params: &pbsubstreams.Module_Input_Params{
+						Value: input.Params,
```
abourget: I don't think this is the right value. We want the parameters from the top-level `Params` mapping in the _Manifest_. That means we'll do a loop on that Manifest:Params and dig into those nested structure to set the `Value` there.
The `input.Params` here would be equal to `"string"` always. But that's not the value we put here. Leave it empty, and loop thorough the top-level manifest's `params` map.

manifest/reader.go
```diff
@@ -282,6 +282,13 @@ func ValidateModules(mods *pbsubstreams.Modules) error {
 
 		for idx, in := range mod.Inputs {
 			switch i := in.Input.(type) {
+			case *pbsubstreams.Module_Input_Params_:
+				if idx != 0 {
+					return fmt.Errorf("module %q: input %d: params must be first input", mod.Name, idx)
+				}
+				if i.Params.Value == "" {
```
abourget: It's perfectly fine that some modules have the empty string for parameters. Just skip this check.

manifest/reader.go
```diff
@@ -415,6 +422,8 @@ func prefixModules(mods []*pbsubstreams.Module, prefix string) {
 				input.Store.ModuleName = prefix + PrefixSeparator + input.Store.ModuleName
 			case *pbsubstreams.Module_Input_Map_:
 				input.Map.ModuleName = prefix + PrefixSeparator + input.Map.ModuleName
+			case *pbsubstreams.Module_Input_Params_:
+				input.Params.ModuleName = prefix + PrefixSeparator + input.Params.ModuleName
```
abourget: Params doesn't need a `ModuleName`. So in the `modules.proto` file:
```
+    message Params {
+      string module_name = 1;
+      string value = 2;
+    }
```
should only be:
```
+    message Params {
+      string value = 1;
+    }
```

The other ones like map and store, are _pointers_ to those modules.. so that's why we need the name. Params are not pointers to any  module, they just are specified or not.

manifest/signature.go
```diff
@@ -74,10 +89,16 @@ func (m *ModuleHashes) HashModule(modules *pbsubstreams.Modules, module *pbsubst
...
+		buf.WriteString(param) //TODO: parse and sort?
+	}
+
 	buf.WriteString("inputs")
 	for _, input := range module.Inputs {
-		buf.WriteString(inputName(input))
-		buf.WriteString(inputValue(input))
+		buf.WriteString(m.inputName(input, module))
+		buf.WriteString(m.inputValue(input, module))
 	}
```
abourget: In this file, the only necessary thing is the addition of the two `case` clauses to add support for `*pbsubstreams.Module_Input_Params_`. The loop through the inputs will cover the params, as well as the value for the params. No need for an additional "params" section, no need for a member `params`.
The params will always be fit _directly_ in the module's `Module_Input_Params_`, and analyzed there.
Did I miss something?

manifest/signature_test.go
```diff
@@ -56,7 +63,10 @@ func Test_HashModule(t *testing.T) {
 	}
 
 	graph, _ := NewModuleGraph(manifest.Modules)
-	hashes := NewModuleHashes()
+	params := map[string]string{
+		"map_pools_created": "foo",
+	}
+	hashes := NewModuleHashes(WithModuleHashesParams(params))
```
abourget: ok I think I understand the confusion. The `Value` in `pbsubstreams.Module_Input_Params` is truly the _actual_ param.. not the type, but the contents of the params. So we don't need to pass them out of band.

pipeline/outputmodules/graph.go
```diff
@@ -120,7 +121,7 @@ func moduleNames(modules []*pbsubstreams.Module) (out []string) {
 }
 
 func (g *Graph) hashModules(graph *manifest.ModuleGraph) {
-	g.moduleHashes = manifest.NewModuleHashes()
+	g.moduleHashes = manifest.NewModuleHashes(manifest.WithModuleHashesParams(params.Decode(g.request.Params)))
```
abourget: ok here again nothing special is needed.. no out-of-band params passing.

pipeline/pipeline.go
```diff
@@ -301,7 +302,7 @@ func (p *Pipeline) buildWASM(ctx context.Context, modules []*pbsubstreams.Module
 	tracer := otel.GetTracerProvider().Tracer("executor")
 
 	for _, module := range modules {
-		inputs, err := p.renderWasmInputs(module)
+		inputs, err := p.renderWasmInputs(module, request)
```
abourget: already passed down! the `module` contains it in the `Input`.

proto/sf/substreams/v1/substreams.proto
```diff
@@ -48,6 +48,8 @@ message Request {
   repeated string debug_initial_store_snapshot_for_modules = 8;
 
   string output_module = 10;
+
+  repeated string params = 11;
```
abourget: no need for this, the `value` is inside the `input`

test/runnable_test.go
```diff
@@ -78,6 +80,10 @@ func (f *testRun) Run(t *testing.T) error {
 		ProductionMode: f.ProductionMode,
 	}
 
+	if f.Params != nil {
+		request.Params = f.Params
+	}
+
```
abourget, line 99: nuke

test/runnable_test.go
```diff
@@ -42,6 +42,8 @@ type testRun struct {
 	ProductionMode         bool
 	Context                context.Context // custom top-level context, defaults to context.Background()
 
+	Params []string // extra params to pass to the substreams service
+
```
abourget, line 47: nuke

pipeline/pipeline.go
```diff
@@ -370,10 +371,18 @@ func returnModuleDataOutputs(clock *pbsubstreams.Clock, step bstream.StepType, c
...
 	for _, input := range module.Inputs {
 		switch in := input.Input.(type) {
+		case *pbsubstreams.Module_Input_Params_:
+			//TODO(colin): is this okay?
+			p := params.Decode(request.Params)
+			inputParams, ok := p[module.Name]
+			if !ok {
+				return nil, fmt.Errorf("params for module %q not found", module.Name)
+			}
+			out = append(out, wasm.NewParamsInput(input.GetParams().GetValue(), inputParams))
```
abourget: Just that last line is needed, and with the `GetValue()`'s value.. no need for `inputParams`. See my comment below on the `NewParamsInput` function.

wasm/arguments.go
```diff
@@ -106,3 +106,19 @@ func NewStoreWriterOutput(name string, store store.Store, updatePolicy pbsubstre
...
+
+func NewParamsInput(name string, value string) *ParamsInput {
+	return &ParamsInput{
+		BaseArgument: BaseArgument{
+			name: name,
+		},
+		BaseValueArgument: BaseValueArgument{
+			value: []byte(value),
+		},
+	}
```
abourget, line 123: That whole `BaseArgument` I think is used only for logging purposes (and even then, if you track it down, that `from string` thing is _not_ even used in `WriteAtPtr()`). So you can make `BaseArgument{name: "params"}` and that'll be fine.
Which means you need only _one_ param here, the `value`.

abourget (commented):
@colindickson my review here :) Good work otherwise! That's going to be cool..
Also, if you can update the docs, that would be a big plus, and the changelog entry.

maoueh:
Seems this has been merged by other means.

PR #171, opened 2023-02-07
Feature/add abigen and events from contract

emiliocramer:


cmd/substreams/init.go
```diff
@@ -39,24 +40,12 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
 
-	// Get desired network
-	networkPrompt := promptui.Select{
-		Label:     "Select network",
-		Items:     []string{"Ethereum", "other"},
-		Templates: choiceTemplate(),
-		HideHelp:  true,
-	}
-	_, network, err := networkPrompt.Run()
+	network, err := generateSelect("Select network", []string{"Ethereum", "other"})
```
maoueh:
```suggestion
	network, err := generateSelect("Select network", []string{"Ethereum", "Other"})
```

cmd/substreams/init.go
```diff
@@ -67,10 +56,47 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
+		contract, err = generatePrompt("Contract to track: ")
+		if err != nil {
+			return fmt.Errorf("running contract prompt: %w", err)
+		}
+	}
+
+	// Clean up given contract
+	contractHex, err := eth.NewHex(contract)
+	if err != nil {
+		return fmt.Errorf("getting contract bytes: %w", err)
```
maoueh: This should be validated while doing `generateSelect("Would you like to track a particular contract", []string{"yes", "no"})`, the library allows for a custom validator

cmd/substreams/init.go
```diff
@@ -67,10 +56,47 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
+	contract := "bc4ca0eda7647a8ab7c2061c2e118a18a936f13d"
+	if wantsAbi == "yes" {
+		contract, err = generatePrompt("Contract to track: ")
+		if err != nil {
+			return fmt.Errorf("running contract prompt: %w", err)
+		}
+	}
+
+	// Clean up given contract
+	contractHex, err := eth.NewHex(contract)
```
maoueh:
```suggestion
	contractAddress, err := eth.NewAddress(contract)
```

cmd/substreams/init.go
```diff
@@ -67,10 +56,47 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return fmt.Errorf("getting contract bytes: %w", err)
+	}
+	contractPretty := contractHex.Pretty()
+
+	// Get contract ABI & parse
+	abi, err := codegen.GetContractAbi(contractPretty)
```
maoueh: You should still use the "better typed" object `contractAddress` when calling `GetContractAbi`, the actual special serialization to `Pretty` that I think it required when you do your HTTP call should be done much closer to where it's required.

You try to work with the correct type, passing it around, until you really need the string version.

Also, Go naming convention is to upper case acronyms so `GetContractABI`

maoueh: I feels like:

```
	ethAbi, err := eth.ParseABIFromBytes(abi)
```

Should be done within `GetContractAbi` directly

cmd/substreams/init.go
```diff
@@ -67,10 +56,47 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
 	fmt.Println("")
 
-	gen := codegen.NewProjectGenerator(srcDir, projectName)
-	if _, err := os.Stat(filepath.Join(srcDir, projectName)); errors.Is(err, os.ErrNotExist) {
+	events, err := codegen.BuildEventModels(ethAbi)
+	if err != nil {
+		return fmt.Errorf("build ABI event models: %w", err)
+	}
+
+	gen := codegen.NewProjectGenerator(srcDir, "new", contract, string(abi), events)
```
maoueh: Why is `"new"` hard-coded here (and in some other places as well)?

codegen/helper.go
```diff
@@ -0,0 +1,159 @@
...
+type CodegenEvent struct {
+	RustName string
+	Fields   map[string]string
+}
+
+func GetContractAbi(contract string) ([]byte, error) {
+	res, err := http.Get(fmt.Sprintf("https://api.etherscan.io/api?module=contract&action=getabi&address=%s&apikey=7E11P1IJ4ZRWQ36CZN78QZ7KTE7YP7MJ31", contract))
+	if err != nil {
+		return nil, fmt.Errorf("getting contract abi from etherscan: %w", err)
+	}
```
maoueh: Are we today showing to the user that it's for `Ethereum Mainnet` only?

codegen/helper.go
```diff
@@ -0,0 +1,159 @@
...
+	"strconv"
+)
+
+type CodegenEvent struct {
+	RustName string
+	Fields   map[string]string
+}
+
+func GetContractAbi(contract string) ([]byte, error) {
+	res, err := http.Get(fmt.Sprintf("https://api.etherscan.io/api?module=contract&action=getabi&address=%s&apikey=7E11P1IJ4ZRWQ36CZN78QZ7KTE7YP7MJ31", contract))
```
maoueh: The actual API key used should probably ce coming from a flag and have the default value. Also, where this API key was generated exactly? And can we use it without one?

codegen/helper.go
```diff
@@ -0,0 +1,159 @@
...
+
+	abi, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, fmt.Errorf("reading fetched abi: %w", err)
+	}
+
+	type Response struct {
+		status  string
+		message string
+		Result  interface{} `json:"result"`
```
maoueh, line 83: Why it's not `string` directly since lower you do `[]byte(response.Result.(string))` which would panic if `interface{}` is not a string?

maoueh: If we keep it `interface{}`, we should check the type before accessing it and present an error message instead of a panicking.

codegen/templates/Cargo.toml.gotmpl
```diff
@@ -16,6 +16,8 @@ num-bigint = "0.4.3"
 ethabi = "17"
 hex-literal = "0.3.4"
 prost = "0.11"
+num-traits = "0.2.15"
+serde_json = "1.0"
```
maoueh: Those will not be required when Protobuf is used directly.

codegen/templates/build.rs.gotmpl
```diff
@@ -0,0 +1,9 @@
+use anyhow::{Ok, Result};
+use substreams_ethereum::Abigen;
+
+fn main() -> Result<(), anyhow::Error> {
+    Abigen::new("Contract", "abi/contractabi.json")?
```
maoueh:
```suggestion
    Abigen::new("Contract", "abi/contract.json")?
```

Feels like enough

maoueh (changes requested):

codegen/helper.go
emiliocramer: it can be done without an API key, changed.

Issue #193, opened 2023-03-14
GUI: error message cut off


YaroShkvorets:
When RPC error is returned, error message (top line) is cut off making it useless

![image](https://user-images.githubusercontent.com/29608734/225129493-e11653fc-7a94-4c8f-b9cc-fb7ee700b095.png)

Should wrap if possible.

PR #194, opened 2023-03-14
Fix error message cutoff

YaroShkvorets:
This should fix an issue with RPC error message cutoff #193
The only way to force word wrap with this library is to set component width explicitly and it depends on the longest label length

Before:

image

After:

Pasted Graphic

### DIFF
--- a/tui2/pages/progress/progress.go
+++ b/tui2/pages/progress/progress.go
@@ -117,5 +117,5 @@ func (p *Progress) SetSize(w, h int) {
 	headerHeight := 7
 	p.Common.SetSize(w, h)
 	p.bars.SetSize(w, h-headerHeight)
-	p.Styles.StatusBarValue.Width(p.Common.Width - labelsMaxLen() - 2) // adjust width to force word wrap
+	p.Styles.StatusBarValue.Width(p.Common.Width - labelsMaxLen()) // adjust status bar width to force word wrap: full width - labels width
 }
--- a/tui2/pages/progress/progress.go
+++ b/tui2/pages/progress/progress.go
@@ -78,13 +78,25 @@ func (p *Progress) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	return p, nil
 }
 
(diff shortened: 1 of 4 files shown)


tui2/pages/progress/progress.go
```diff
@@ -105,4 +105,5 @@ func (p *Progress) SetSize(w, h int) {
 	headerHeight := 7
 	p.Common.SetSize(w, h)
 	p.bars.SetSize(w, h-headerHeight)
+	p.Styles.StatusBarValue.Width(p.Common.Width - len("Parallel engine progress messages: ") - 2)	// adjust width to force word wrap
```
maoueh: Shouldn't we directly have a constant named and properly documented for the `len("Parallel engine progress messages: ") - 2` part?

YaroShkvorets: To avoid magic numbers I've added a helper function to calculate labels pane width in case labels change in the future.
If you prefer to just hardcode the number at 36 let me know.

tui2/pages/progress/progress.go
```diff
@@ -105,4 +117,5 @@ func (p *Progress) SetSize(w, h int) {
 	headerHeight := 7
 	p.Common.SetSize(w, h)
 	p.bars.SetSize(w, h-headerHeight)
+	p.Styles.StatusBarValue.Width(p.Common.Width - labelsMaxLen() - 2) // adjust width to force word wrap
```
maoueh: Seems good, can you just document the reason for the `-2`.

Update the CHANGELOG.md file and we are good to go.

maoueh (changes requested):

tui2/pages/progress/progress.go
YaroShkvorets: `2` is for padding. Moved it to the `styles.go` since it doesn't change.
Updated changelog.

YaroShkvorets:
Fixed a typo from an unrelated commit that was failing one of the tests

PR #204, opened 2023-04-06
Rework undo

sduchesneau:
* Services for tier1 (request, response...) now exposed as `sf.substreams.rpc.v2.Stream/Blocks` and removed from `sf.substreams.v1`
  * Single module output from mapper in response. 
  * In dev mode, the additional modules output can be inspected under `debug_map_outputs` and `debug_store_outputs` 
  * The concept of "step" under blockData has been removed: reorgs are now signalled with a `BlockUndoSignal` message (without the original data)
* Services for tier2 (subrequests) now exposed as `sf.substreams.internal.v2.Substreams/ProcessRange` (not to be used externally)
* The package and modules definitions stay under `sf.substreams.v1`
* If a cursor cannot be resolved to the closest canonical parent, it will fall back to the closest final block.


sduchesneau:
```substreams run -e api-unstable.streamingfast.io:443  gs://staging.dfuseio-global.appspot.com/substreams/eth-block-meta/substreams-eth-block-meta-sync-v0.4.0.spkg  blockmeta:kv_out  --cursor=$(opaque encode c1:1:16926830:af603ab44685158e99687887c57db0081fd12476299f44ebb26f165f7739072d:16926739:2c4d685eabea79dc39be23e90f101b18713922a9e54d7fc5dc7ecf9a3eec3765) -t 16926832```
example getting out of a bad cursor ^

File: docs/release-notes/change-log.md [hunk 1]
```diff
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 1.1.0 (Unreleased)
+
+### Highlights -- BREAKING CHANGES
```
maoueh: We should have a section per for RPC communication and sub-section for Undo signal change. We should use a more doc-centric approach with longer description of the change, what it affects and upgrade path, we could should "trimmed down" Golang code for consuming Substreams before/after.

docs/release-notes/change-log.md
```diff
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
...
+
+* Services for tier1 (request, response...) now exposed as `sf.substreams.rpc.v2.Stream/Blocks` and removed from `sf.substreams.v1`
+  * Single module output from mapper in response. 
+  * In dev mode, the additional modules output can be inspected under `debug_map_outputs` and `debug_store_outputs`
```
maoueh: Single module output was already enforced but it's now removed from the protocol, we should just explained it's removal of deprecated fields, to adjust the tone a bit.

docs/release-notes/change-log.md
```diff
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
...
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 1.1.0 (Unreleased)
+
+### Highlights -- BREAKING CHANGES
+
+* Services for tier1 (request, response...) now exposed as `sf.substreams.rpc.v2.Stream/Blocks` and removed from `sf.substreams.v1`
+  * Single module output from mapper in response. 
+  * In dev mode, the additional modules output can be inspected under `debug_map_outputs` and `debug_store_outputs` 
+  * The concept of "step" under blockData has been removed: reorgs are now signalled with a `BlockUndoSignal` message (without the original data)
```
maoueh: Would go in undo sub-section of RPC highlights with more meat about what change exactly.

docs/release-notes/change-log.md
```diff
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
...
+* Services for tier2 (subrequests) now exposed as `sf.substreams.internal.v2.Substreams/ProcessRange` (not to be used externally)
```
maoueh: I would personally talk in a dedicated operators section

manifest/manifest_test.go
```diff
@@ -4,11 +4,10 @@ import (
 	"strings"
 	"testing"
 
+	pbsubstreams "github.com/streamingfast/substreams/pb/sf/substreams/v1"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
+	"github.com/test-go/testify/require"
```
maoueh: Wrong test package

orchestrator/parallelprocessor.go
```diff
@@ -32,10 +35,12 @@ func BuildParallelProcessor(
 	runtimeConfig config.RuntimeConfig,
 	outputGraph *outputmodules.Graph,
 	execoutStorage *execout.Configs,
-	respFunc func(resp *pbsubstreams.Response) error,
+	respFunc func(resp substreams.ResponseFromAnyTier) error,
 	storeConfigs store.ConfigMap,
 ) (*ParallelProcessor, error) {
 	var execOutputReader *execout.LinearReader
+	_ = pbssinternal.ProcessRangeRequest{}
+	_ = pbsubstreamsrpc.Request{}
```
maoueh: Leftover?

client/client.go
```diff
@@ -8,7 +8,8 @@ import (
 	"regexp"
 
 	"github.com/streamingfast/dgrpc"
-	pbsubstreams "github.com/streamingfast/substreams/pb/sf/substreams/v1"
+	pbssinternal "github.com/streamingfast/substreams/pb/sf/substreams/intern/v2"
```
maoueh, line 11: Not sure why we are trying to cut `al` in `intern`within `sf/substreams/intern/v2`, later we also use `pbssintern` for module and it's confusing. 

We should stick with `sf/substreams/internal/v2` and `pbssinternal`.

proto/sf/substreams/v1/test/test.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
 
-package sf.substreams.v1.test;
-option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/v1/test/;pbsubstreamstest";
+package sf.substreams.v2.test;
+option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/v2/test/;pbsubstreamstest";
```
maoueh: Why changing the version here to `v2`?

service/tier1.go
```diff
@@ -252,31 +236,29 @@ func (s *Service) blocks(ctx context.Context, runtimeConfig config.RuntimeConfig
...
 		return fmt.Errorf("configuring stores: %w", err)
 	}
-	stores := pipeline.NewStores(storeConfigs, runtimeConfig.CacheSaveInterval, requestDetails.RequestStartBlockNum, request.StopBlockNum, isSubRequest)
+	stores := pipeline.NewStores(storeConfigs, runtimeConfig.CacheSaveInterval, requestDetails.RequestStartBlockNum, request.StopBlockNum, false)
 
 	// TODO(abourget): why would this start at the LinearHandoffBlockNum ?
 	//  * in direct mode, this would mean we start writing files after the handoff,
 	//    but it's not so useful to write those files, as they're partials
 	//    and the OutputWriter doesn't know if that `initialBlockBoundary` is the  module's init Block?
 	//  *
```
maoueh: Seems to me like this was attached to `execOutWriter` code block that existed here before. Should be moved with code or removed.

proto/sf/substreams/rpc/v2/clock.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
 
-package sf.substreams.v1;
-option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/v1;pbsubstreams";
+package sf.substreams.rpc.v2;
```
maoueh: The `Clock` cannot moved, it's used in `substreams.yaml` as input type for some module, we must keep it under `sf.substreams.v1`

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -39,43 +39,51 @@ message Request {
...
 
+
 message Response {
   oneof message {
-    SessionInit session = 5; // Always sent first
-    ModulesProgress progress = 1; // Progress of data preparation, before sending in the stream of `data` events.
+    SessionInit session = 1; // Always sent first
+    ModulesProgress progress = 2; // Progress of data preparation, before sending in the stream of `data` events.
+    BlockScopedData block_data = 3; 
+    BlockUndoSignal block_undo_signal = 4;
```
maoueh, line 58: Should we think about adding `BlockFinalSignal` message right now? This could be useful for implementation where a simple "fork buffer" could be kept more easily.

We can safely add in a future update also, so maybe post-pone.

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -39,43 +39,51 @@ message Request {
...
-  // Removed, was STEP_REDO
-  reserved 3;
-  // Block is now irreversible and can be committed to (finality is chain specific, see chain documentation for more details)
-  STEP_IRREVERSIBLE = 4;
-  // Removed, was STEP_STALLED
-  reserved 5;
+
+// BlockUndoSignal informs you that every bit of data
+// with a clock number above 'last_valid_clock' has been reverted
+// on-chain. Delete that data and restart from 'last_valid_cursor'
```
maoueh: Can you add an example using example block nums for the "flow" and what should happen when handling it.

maoueh: Should we think about adding `BlockFinalSignal` message right now? This could be useful for implementation where a simple "fork buffer" could be kept more easily.

After reading more code, it seems we have `final_block_height` within the `BlockScopedData`, so it really raise the question about `BlockFinalSignal`.

orchestrator/work/worker.go
```diff
@@ -148,56 +148,101 @@ func (w *RemoteWorker) Work(ctx context.Context, request *pbsubstreams.Request,
...
-					if f := progress.GetProcessedBytes(); f != nil {
-						bm := tracking.GetBytesMeter(ctx)
-						bm.AddBytesWritten(int(f.TotalBytesWritten))
-						bm.AddBytesRead(int(f.TotalBytesRead))
-					}
+			case *pbssinternal.ProcessRangeResponse_ProcessedBytes:
+				//		bm := tracking.GetBytesMeter(ctx)
+				//		bm.AddBytesWritten(int(f.TotalBytesWritten))
+				//		bm.AddBytesRead(int(f.TotalBytesRead))
+				// respFunc(...)
```
maoueh: What should we do with those dead code? If we keep them, they should be updated to compile with code and then we comment them out so at least they are aligned with current code.

orchestrator/work/worker.go
```diff
@@ -148,56 +148,101 @@ func (w *RemoteWorker) Work(ctx context.Context, request *pbsubstreams.Request,
...
 			}
 		}
 	}
 }
+func toRPCFailedProgressResponse(moduleName, reason string, logs []string, logsTruncated bool) *pbsubstreamsrpc.Response {
+	return &pbsubstreamsrpc.Response{
+		Message: &pbsubstreamsrpc.Response_Progress{
+			Progress: &pbsubstreamsrpc.ModulesProgress{
+				Modules: []*pbsubstreamsrpc.ModuleProgress{
+					&pbsubstreamsrpc.ModuleProgress{
```
maoueh:
```suggestion
					{
```

orchestrator/work/worker.go
```diff
@@ -148,56 +148,101 @@ func (w *RemoteWorker) Work(ctx context.Context, request *pbsubstreams.Request,
...
+		},
+	}
+}
+
+func toRPCRangeProgressResponse(moduleName string, start, end uint64) *pbsubstreamsrpc.Response {
+	return &pbsubstreamsrpc.Response{
+		Message: &pbsubstreamsrpc.Response_Progress{
+			Progress: &pbsubstreamsrpc.ModulesProgress{
+				Modules: []*pbsubstreamsrpc.ModuleProgress{
+					&pbsubstreamsrpc.ModuleProgress{
```
maoueh: Same here 
```suggestion
					{
```

pipeline/outputmodules/graph.go
```diff
@@ -8,11 +8,10 @@ import (
 )
 
 type Graph struct {
-	request *pbsubstreams.Request
-
-	allModules   []*pbsubstreams.Module // all modules that need to be processed (requested directly or a required module ancestor)
-	moduleHashes *manifest.ModuleHashes
-	stores       []*pbsubstreams.Module // subset of allModules: only the stores
+	requestModules *pbsubstreams.Modules
+	usedModules    []*pbsubstreams.Module // all modules that need to be processed (requested directly or a required module ancestor)
```
maoueh, line 12: Excellent renamed!

maoueh, line 12: Excellent rename!

pipeline/outputmodules/graph.go
```diff
@@ -22,44 +21,44 @@ type Graph struct {
 
 func (g *Graph) OutputModule() *pbsubstreams.Module   { return g.outputModule }
 func (g *Graph) Stores() []*pbsubstreams.Module       { return g.stores }
-func (g *Graph) AllModules() []*pbsubstreams.Module   { return g.allModules }
+func (g *Graph) AllModules() []*pbsubstreams.Module   { return g.usedModules }
```
maoueh: Should be renamed too.

pipeline/outputmodules/validate.go
```diff
@@ -4,11 +4,13 @@ import (
...
 
 	"github.com/streamingfast/substreams/manifest"
+	pbssinternal "github.com/streamingfast/substreams/pb/sf/substreams/intern/v2"
+	pbsubstreamsrpc "github.com/streamingfast/substreams/pb/sf/substreams/rpc/v2"
 	pbsubstreams "github.com/streamingfast/substreams/pb/sf/substreams/v1"
 )
 
 // ValidateRequest is run by the server code.
-func ValidateRequest(request *pbsubstreams.Request, blockType string, isSubRequest bool) error {
+func ValidateRequest(request *pbsubstreamsrpc.Request, blockType string, isSubRequest bool) error {
```
maoueh: `isSubRequest` is unused.

pipeline/outputmodules/validate.go
```diff
@@ -35,10 +37,72 @@ func ValidateRequest(request *pbsubstreams.Request, blockType string, isSubReque
...
+		return fmt.Errorf("computing ancestors of %q: %w", request.OutputModule, err)
+	}
+
+	// We must only validate the input source against module that we are going to actually run. A Substreams
+	// could provide modules for multiple chain while executing only one of them in which case only the one
+	// run (and its dependencies transitively) should be checked.
+	for _, mod := range ancestors {
+		for _, input := range mod.Inputs {
+			if src := input.GetSource(); src != nil {
+				if src.Type != blockType && src.Type != "sf.substreams.v1.Clock" {
```
maoueh: Seems another good readon to keep the clock at `sf.substreams.v1.Clock`

pipeline/outputmodules/validate.go
```diff
@@ -35,10 +37,72 @@ func ValidateRequest(request *pbsubstreams.Request, blockType string, isSubReque
...
+					return fmt.Errorf("input source %q not supported, only %q and 'sf.substreams.v1.Clock' are valid", src, blockType)
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func ValidateInternalRequest(request *pbssinternal.ProcessRangeRequest, blockType string) error {
```
maoueh, line 33: You have put the non internal method directly under the Request type receiver, should we do the name here?

pipeline/init_test.go
```diff
@@ -59,7 +59,7 @@ func NewExecOutputTesting(t *testing.T, block *bstream.Block, clock *pbsubstream
 		clock: clock,
 		Values: map[string][]byte{
 			"sf.substreams.v1.test.Block": blkBytes,
-			"sf.substreams.v1.Clock":      clockBytes,
+			"sf.substreams.rpc.v2.Clock":  clockBytes,
```
maoueh: To be reverted.

pipeline/pipeline.go
```diff
@@ -93,7 +102,7 @@ func (p *Pipeline) InitStoresAndBackprocess(ctx context.Context) (err error) {
 	ctx, span := reqctx.WithSpan(ctx, "pipeline_init")
 	defer span.EndWithErr(&err)
 
-	p.forkHandler.registerUndoHandler(func(clock *pbsubstreams.Clock, moduleOutputs []*pbsubstreams.ModuleOutput) {
+	p.forkHandler.registerUndoHandler(func(clock *pbsubstreamsrpc.Clock, moduleOutputs []*pbssinternal.ModuleOutput) {
```
maoueh: `clock` is not used anymore, should the undo handler be modified?

pipeline/pipeline.go
```diff
@@ -354,16 +454,23 @@ func (p *Pipeline) buildWASM(ctx context.Context, modules []*pbsubstreams.Module
...
+	out := &pbsubstreamsrpc.BlockScopedData{
+		Clock:             clock,
+		Output:            mapModuleOutput,
+		DebugMapOutputs:   extraMapModuleOutputs,
+		DebugStoreOutputs: extraStoreModuleOutputs,
+		Cursor:            cursor.ToOpaque(),
 	}
 
-	if err := respFunc(substreams.NewBlockScopedDataResponse(out)); err != nil {
+	if err := respFunc(substreams.NewBlockDataResponse(out)); err != nil {
```
maoueh: Not sure why we renamed, maybe you would like to have `BlockData` instead in the Proto def :)?

pipeline/process_block.go
```diff
@@ -58,18 +56,42 @@ func (p *Pipeline) ProcessBlock(block *bstream.Block, obj interface{}) (err erro
...
+		Id:        block.Id,
+		Timestamp: timestamppb.New(block.Time()),
+	}
+}
+
+func blockRefToClock(block bstream.BlockRef) *pbsubstreamsrpc.Clock {
+	return &pbsubstreamsrpc.Clock{
+		Number: block.Num(),
+		Id:     block.ID(),
+		// Timestamp: timestamppb.New(block.Time()),
```
maoueh: Hmmm, this makes me wonder that maybe we should use `BlockRef` instead of `Clock` in the RPC layer where this is used, maybe just in the `undo` case. If we are not able to use the `Timestamp` everywhere, we need to seriously think about using a different type.

pipeline/process_block.go
```diff
@@ -112,52 +134,65 @@ func (p *Pipeline) processBlock(ctx context.Context, block *bstream.Block, clock
...
+
+	return p.respFunc(
+		&pbsubstreamsrpc.Response{
+			Message: &pbsubstreamsrpc.Response_BlockUndoSignal{
+				&pbsubstreamsrpc.BlockUndoSignal{
+					LastValidClock:  targetClock,
+					LastValidCursor: targetCursor.ToOpaque(),
+				},
+			},
+		})
```
maoueh:
```suggestion
	return p.respFunc(
		&pbsubstreamsrpc.Response{
			Message: &pbsubstreamsrpc.Response_BlockUndoSignal{
				BlockUndoSignal: &pbsubstreamsrpc.BlockUndoSignal{
					LastValidClock:  targetClock,
					LastValidCursor: targetCursor.ToOpaque(),
				},
			},
		})
```

pipeline/process_block.go
```diff
@@ -179,14 +214,26 @@ func (p *Pipeline) handlerStepNew(ctx context.Context, block *bstream.Block, clo
...
+			}
 		}
 	}
 
 	if p.gate.shouldSendOutputs() {
 		logger.Debug("will return module outputs")
-		if err = returnModuleDataOutputs(clock, bstream.StepNew, cursor, p.moduleOutputs, p.respFunc); err != nil {
+		for _, hook := range p.preFirstBlockDataHooks {
+			if err := hook(ctx, clock); err != nil {
+				return fmt.Errorf("failed to run pre-first-block-data-hook: %w", err)
```
maoueh:
```suggestion
				return fmt.Errorf("failed to run pre first block data hook: %w", err)
```

I think we are safe to use the spaced version.

pipeline/process_block.go
```diff
@@ -179,14 +214,26 @@ func (p *Pipeline) handlerStepNew(ctx context.Context, block *bstream.Block, clo
...
+			}
+		}
+		p.preFirstBlockDataHooks = nil
```
maoueh, line 235: I'm a bit skeptical about using this to use it as the "first has been executed" way of things.

pipeline/resolve.go
```diff
@@ -83,23 +103,87 @@ func resolveStartBlockNum(req *pbsubstreams.Request, getHeadBlock getBlockFunc)
...
 	if err != nil {
-		return 0, status.Errorf(grpccodes.InvalidArgument, "invalid start cursor %q: %s", cursor, err.Error())
+		return 0, nil, status.Errorf(grpccodes.InvalidArgument, "invalid start cursor %q: %s", cursor, err.Error())
 	}
-	if cursor.Step.Matches(bstream.StepIrreversible) {
-		return cursor.Block.Num() + 1, nil // this block was the last sent to the customer
+
+	if cursor.IsOnFinalBlock() {
+		nextBlock := cursor.Block.Num() + 1
+		req.StartCursor = ""
```
maoueh: Are we loosing some "visibility" on the real actual request by doing this? Could we instead find a way to keep the request StartCursor but use a "resolvedCursor"?

pipeline/resolve.go
```diff
@@ -83,23 +103,87 @@ func resolveStartBlockNum(req *pbsubstreams.Request, getHeadBlock getBlockFunc)
...
+		return nextBlock, nil, nil
 	}
-	if cursor.Step.Matches(bstream.StepNew) {
-		return cursor.Block.Num() + 1, nil // this block was the last sent to the customer
+
+	lastValidBlock, head, err := resolveCursor(ctx, cursor)
+	if err != nil {
+		return 0, nil, status.Errorf(grpccodes.InvalidArgument, "cannot resolve startCursor %q: %s", cursor, err.Error())
```
maoueh: In the other cursor related error, we use `start cursor`, I would be in favor of doing this instead:
```suggestion
		return 0, nil, status.Errorf(grpccodes.InvalidArgument, "cannot resolve StartCursor %q: %s", cursor, err.Error())
```

And normalize the other one above. This way, the user see immediately which field is the problem.

pipeline/resolve.go
```diff
@@ -83,23 +103,87 @@ func resolveStartBlockNum(req *pbsubstreams.Request, getHeadBlock getBlockFunc)
...
-		return cursor.Block.Num(), nil
+	var undoSignal *pbsubstreamsrpc.BlockUndoSignal
+	if lastValidBlock.Num() != cursor.Block.Num() {
+		validCursor := &bstream.Cursor{
+			Step:      bstream.StepNew,
+			Block:     lastValidBlock,
+			LIB:       cursor.LIB,
+			HeadBlock: head,
+		}
+		req.StartCursor = validCursor.ToOpaque()
```
maoueh: Same thing here about modified the origin request.

pipeline/resolve.go
```diff
@@ -83,23 +103,87 @@ func resolveStartBlockNum(req *pbsubstreams.Request, getHeadBlock getBlockFunc)
...
+
+func NewCursorResolver(hub *hub.ForkableHub, mergedBlocksStore, forkedBlocksStore dstore.Store) CursorResolver {
+
+	return func(ctx context.Context, cursor *bstream.Cursor) (lastValidBlock, currentHead bstream.BlockRef, err error) {
+		jctBlkGetter := &junctionBlockGetter{}
+		src := hub.SourceFromCursor(cursor, jctBlkGetter)
+		if src == nil { // block is out of reversible segment
+			src = bstream.NewFileSourceFromCursor(mergedBlocksStore, forkedBlocksStore, cursor, jctBlkGetter, zap.NewNop())
+		}
+		if src == nil {
```
maoueh: `src` cannot be `nil` here, the previous condition force it to be `non-nil`.

maoueh: I verified, `NewFileSourceFromCursor` seems it cannot return `nil`.

pipeline/resolve_test.go
```diff
@@ -36,8 +39,8 @@ func Test_resolveStartBlockNum(t *testing.T) {
 			wantErr:          true,
 		},
 		{
-			name: "step undo",
-			req: &pbsubstreams.Request{
+			name: "step undo", // substreams customers should not receive these cursors now, kept for backwards compatibility
```
maoueh: I'm not sure what backward compatibility we are keeping here.

service/tier2.go
```diff
@@ -0,0 +1,279 @@
...
+	outputGraph, err := outputmodules.NewOutputModuleGraph(request.OutputModule, true, request.Modules)
+	if err != nil {
+		return stream.NewErrInvalidArg(err.Error())
+	}
+
+	ctx, requestStats := setupRequestStats(ctx, logger, runtimeConfig.WithRequestStats, true)
+
+	//bytesMeter := tracking.NewBytesMeter(ctx)
+	//bytesMeter.Launch(ctx, respFunc)
+	//ctx = tracking.WithBytesMeter(ctx, bytesMeter)
```
maoueh: To be updated?

service/tier1.go
maoueh: Seems that it should be deleted because it's now in `Tier2Service`

storage/store/value_get.go
```diff
@@ -135,13 +135,13 @@ func (b *baseStore) HasAt(ord uint64, key string) bool {
...
-		case pbsubstreams.StoreDelta_DELETE, pbsubstreams.StoreDelta_UPDATE:
+		case pbssintern.StoreDelta_DELETE, pbssintern.StoreDelta_UPDATE:
 			found = true
-		case pbsubstreams.StoreDelta_CREATE:
+		case pbssintern.StoreDelta_CREATE:
 			found = false
 		default:
 			// WARN: is that legit? what if some upstream stream is broken? can we trust all those streams?
-			panic(fmt.Sprintf("invalid value %q for pbsubstreams.StateDelta::Op for key %q", delta.Operation, delta.Key))
+			panic(fmt.Sprintf("invalid value %q for pbsspriv.StateDelta::Op for key %q", delta.Operation, delta.Key))
```
maoueh: All occurrences of `pbsspriv` should become `pbssinternal`

maoueh (changes requested):
Main thing is around the Clock that I think should not move, there is some other stuff here and there.

maoueh (changes requested):
Main thing is around the Clock that I think should not move, there is some other stuff here and there.

maoueh (changes requested):
Main thing is around the Clock that I think should not move, there is some other stuff here and there.

docs/release-notes/change-log.md
```diff
@@ -4,6 +4,50 @@ All notable changes to this project will be documented in this file.
...
+Now that the `Blocks` request has been moved from `sf.substreams.v1` to
+`sf.substreams.rpc.v2`, the communication between a substreams instance acting
+as tier1 and a tier2 instance that performs the background processing has also
+been reworked, and put under `sf.substreams.internal.v2.Stream/ProcessRange`.
+It has also been stripped of parameters that were not used for that level of
+communication (ex: `cursor`, `logs`...)
+
+### Other changes in protobuf messages
+
+* Deprecating the "Request" and "Response" messages, with some others, from `sf.substreams.v1`, now that they moved to `sf.substreams.rpc.v2`
```
maoueh, line 46: Deprecating or removing?

maoueh (approved):

PR #223, opened 2023-05-26
chore: add github workflow for buf linting & push

fubhy:
Something like that @maoueh?

Since you guys are using github releases anyways, why don't you use the `release` trigger for your workflows to tag docker images, gitpod images & this? That would imho be a better match.

There are currently a few linting errors in `buf lint`. Do we/you care about those? Do we want to fix them?

### DIFF
--- a/.github/workflows/buf.yml
+++ b/.github/workflows/buf.yml
@@ -1,0 +1,42 @@
+name: Buf
+
+on:
+  pull_request:
+  push:
+    tags:
+      - v*
+    branches:
+      - develop
+      - release/v*
+      - feature/*
+
+jobs:
+  buf:
+    runs-on: ubuntu-20.04
+    if: "${{ !startsWith(github.event.head_commit.message, 'GitBook: [#') }}"
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup buf
+        uses: bufbuild/buf-setup-action@v1
+
+      - name: Lint protos
+        # TODO: Consider enabling this after fixing or ignoring the current linting errors?
+        if: false
+        uses: bufbuild/buf-lint-action@v1
+          
+      - name: Check for breaking changes
+        if: github.event_name == 'pull_request'
+        uses: bufbuild/buf-breaking-action@v1
+        with:
+          against: 'https://github.com/${{ github.repository }}.git#branch=${{ github.base_ref }}'
+
+      - uses: bufbuild/buf-push-action@v1
+        if: github.event_name != 'pull_request'
+        with:
+          buf_token: ${{ secrets.BUF_TOKEN }}
+          draft: ${{ !startsWith(github.ref, 'refs/tags/v') }}


maoueh:
I'm not against `release` but I always feel that tags are more useful with similar flow. Main diff is that it enables doing ad-hoc tag, but I could agree it's limited (we could still do `-rc`, `-beta` release).

I leave you the final decision, I'm good with both. Tell me.

maoueh:
The rest is all good.

fubhy:
> The rest is all good.

Let's stick to the current branch, pr & tags model for now then. If we feel like there are signfiicant benefits to using `release` fully in the future across all your repos then we can do that for all of them in one go. I think it's better to just keep them all aligned for now. Less confusing.

Once you merge this, I'll open PRs for all the sinks, other repos, etc. (anything substreams related) that host `proto` files.

.github/workflows/buf.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Check for breaking changes
+        if: github.event_name == 'pull_request'
+        uses: bufbuild/buf-breaking-action@v1
+        with:
+          against: 'https://github.com/${{ github.repository }}.git#branch=${{ github.base_ref }}'
+
+      - uses: bufbuild/buf-push-action@v1
+        if: github.event_name != 'pull_request'
+        with:
+          buf_token: ${{ secrets.BUF_TOKEN }}
```
fubhy, line 41: We need to set this secret in the repo or organization config.

.github/workflows/buf.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup buf
+        uses: bufbuild/buf-setup-action@v1
+
+      - name: Lint protos
+        # TODO: Consider enabling this after fixing or ignoring the current linting errors?
+        if: false
```
fubhy, line 29: Linting is currently disabled because there are some linting errors already. If you want to enforce linting (according to the `buf` rules that is), we can enable this later.

.github/workflows/buf.yml
maoueh: Done

.github/workflows/buf.yml
maoueh: If will take a look but yes, I'm in for having proper linting enabled.

pistomat:
What is the update on the fix of linting? We are using linting in a repository that is dependent on this one and it forces us to have linting on substreams disabled or maintain a fork of the code. 

I can create a PR if there is enough demand. Would renaming of folder `intern` to `internal` as per the docs mean breaking changes for depending libs?

cc @maoueh @fubhy

maoueh:
Sadly we cannot rename `intern` to `internal`. The problem with this one is that `internal` is a special "module" in Golang which prevents other from importing it that's why we needed to name the package `intern`. I'm surprised though that it create a lint warning/error.

About the other lint, would need to evaluate them. Do you know if they have "suppress warnings" that we could maybe directly add to the proto source?

Issue #247, opened 2023-07-06
`alpha init`: set `initialBlock`


YaroShkvorets:
Generator could pull contract creation block from etherscan along with ABI and set `initialBlock` in `substreams.yaml` instead of using `0`: https://github.com/streamingfast/substreams/blob/faed57f78173422335427092dbe5ecb6b1ae82de/codegen/templates/ethereum/substreams.yaml#L20

maoueh:
@YaroShkvorets In your PR that fixes an issue, add in the PR's description `Fixes #<bug>`, GitHub will auto-close the issue on merge.

PR #249, opened 2023-07-06
`init`: set initialBlock in generated manifest

YaroShkvorets:
ref: #247 

1. request contract creation transaction from etherscan
2. parse it
3. set initialBlock in generated `substreams.yaml`

### DIFF
--- a/cmd/substreams/init.go
+++ b/cmd/substreams/init.go
@@ -99,6 +99,15 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return fmt.Errorf("getting contract ABI: %w", err)
 		}
+		// Wait 5 seconds to avoid Etherscan API rate limit
+		time.Sleep(5 * time.Second)
+
+		// Get contract creation block
+		creationBlockNum, err := GetContractCreationBlock(cmd.Context(), contract)
+		if err != nil {
+			fmt.Printf("Failed getting contract creation block, using 0 instead: %v", err)
+			creationBlockNum = "0"
+		}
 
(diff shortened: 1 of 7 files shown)


cmd/substreams/init.go
```diff
@@ -401,3 +411,35 @@ func GetContractABI(ctx context.Context, contract eth.Address) (string, *eth.ABI
 
 	return abiContent, ethABI, nil
 }
+
+func GetContractCreationBlock(ctx context.Context, contract eth.Address) (string, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("https://api.etherscan.io/api?module=account&action=txlist&address=%s&page=1&offset=1&sort=asc&apikey=YourApiKeyToken", contract.Pretty()), nil)
```
DenisCarriere, line 416: @YaroShkvorets since `apikey` isn't being replaced, could be remove, the 5 call/second would still be applied without providing API key.

Also, would it be possible to include the other Etherscan networks.

- https://api-goerli.etherscan.io/api
- https://api-sepolia.etherscan.io/api

There's some more as well such as BNB & Polygon:
- https://api.bscscan.com/api
- https://api.polygonscan.com/api

YaroShkvorets: Yeah currently generator works for Ethereum mainnet.

image

Ideally the wizard should ask for the network (Mainnet, BNB, Polygon, Testnets) after selecting the protocol. But this is a separate issue.

maoueh:
Missing CHANGELOG entry can you open a separate PR, sorry I missed it.

Issue #255, opened 2023-07-11
`alpha init`: prompt for Ethereum chain


YaroShkvorets:
Right now `alpha init` only works for Ethereum Mainnet.

Should prompt for chain: Mainnet, Goerli, Sepolia, BNB, Polygon and generate corresponding code

PR #256, opened 2023-07-11
`alpha init`: prompt chain

YaroShkvorets:
Implements #255

- If Ethereum protocol selected - prompt chain
- Mainnet/BNB/Polygon/Goerli/Sepolia available


codegen/templates/ethereum/Makefile
```diff
@@ -1,18 +1,16 @@
...
-STOP_BLOCK ?= +100
 
 .PHONY: build
 build:
 	cargo build --target wasm32-unknown-unknown --release
 
 .PHONY: run
 run: build
-	substreams run -e $(ENDPOINT) substreams.yaml map_events -s $(START_BLOCK) -t $(STOP_BLOCK)
+	substreams run -e $(ENDPOINT) substreams.yaml map_events $(if $(START_BLOCK),-s $(START_BLOCK)) $(if $(STOP_BLOCK),-t $(STOP_BLOCK))
```
maoueh, line 9: Nice

codegen/templates/ethereum_chain.go
```diff
@@ -1,15 +1,55 @@
...
+		DefaultContractName:    "GETH Token",
+		FirehoseEndpoint:       "goerli.eth.streamingfast.io:443",
+	},
+	"Sepolia": {
+		DisplayName:            "Sepolia Testnet",
+		ExplorerLink:           "https://sepolia.etherscan.io",
+		ApiEndpoint:            "https://api-sepolia.etherscan.io",
+		DefaultContractAddress: "0x1f9840a85d5aF5bf1D1762F925BDADdC4201F984",
+		DefaultContractName:    "UNI Token",
+		FirehoseEndpoint:       "sepolia.streamingfast.io:443",
```
maoueh: Can you temporary comment this one, we don't have an endpoint available for this yet.

maoueh (approved):
Approving because everything else is perfect!

codegen/templates/ethereum_chain.go
YaroShkvorets: I can use Pinax endpoint for Sepolia if that's OK with you: sepolia.firehose.pinax.network:9000

maoueh: I thought about this but the auth is different and this is not documented today, so for now, let's just keep it commented out.

YaroShkvorets: Removed Sepolia

PR #266, opened 2023-07-21
modify proto (rpc+internal) to allow more module/jobs stats

sduchesneau:
(WIP)

Tier1 should aggregate data from tier2 jobs and modules, send stats back to the client


proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -130,6 +130,37 @@ message OutputDebugInfo {
 
 message ModulesProgress {
   repeated ModuleProgress modules = 1;
+  repeated Job running_jobs = 2;
+  repeated ModuleStats modules_stats = 3;
+}
+
+message Job {
```
maoueh, line 163: If we move forward, we need to document all this field properly. That will be important especially on `ModuleStats` to correctly describe the "spec" and what each metric means exactly as well a potential gotchas.

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -130,6 +130,37 @@ message OutputDebugInfo {
...
+    uint64 processed_blocks = 5;
+    uint64 duration_secs = 6;
+}
+
+message ModuleStats {
+    string name = 1;
+
+    // total_processed_blocks includes duplicate for each stage going over that module
+    uint64 total_processed_blocks = 2;
+    uint64 highest_processed_block = 3;
```
maoueh: Is this globally resolved? In the sense that it will always report the highest block of the "unified" view of the module and not just for the current run?

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -130,6 +130,37 @@ message OutputDebugInfo {
...
+    uint64 total_processing_time_secs = 4;
+    uint64 total_external_calls_time_secs = 5;
```
maoueh: What is external calls? Like `rpc` calls in general? Or intrinsics in general?

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -130,6 +130,37 @@ message OutputDebugInfo {
...
+message Job {
+    uint64 stage = 1;
+    repeated string modules = 2;
+    uint64 start_block = 3;
+    uint64 stop_block = 4;
+    uint64 processed_blocks = 5;
+    uint64 duration_secs = 6;
+}
+
+message ModuleStats {
```
maoueh, line 178: What will be the frequency of report? One per module chain segment?

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -130,6 +130,37 @@ message OutputDebugInfo {
...
+    string name = 1;
+
+    // total_processed_blocks includes duplicate for each stage going over that module
+    uint64 total_processed_blocks = 2;
+    uint64 highest_processed_block = 3;
+    uint64 total_processing_time_secs = 4;
+    uint64 total_external_calls_time_secs = 5;
+    uint64 total_external_calls_count = 6;
+    uint64 total_store_operations_time_secs = 7;
```
maoueh: I imagine we will be more in `ms` and less no?

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -130,6 +130,37 @@ message OutputDebugInfo {
...
+    uint64 total_store_reads_count = 8;
+    uint64 total_errors_count = 9;
```
maoueh: What is this, seems weird as a progress, what do you see as being tracked here, retry of a job?

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -130,6 +130,37 @@ message OutputDebugInfo {
...
+
+    // store-specific (will be 0 on mappers)
```
maoueh: Maybe a one of for each different kind

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -130,6 +130,37 @@ message OutputDebugInfo {
...
+    uint64 total_store_writes_count = 10;
+    uint64 total_store_deleteprefix_count = 11;
+    uint64 store_size_bytes = 12;
+    uint64 highest_merged_block = 13;
```
maoueh: Not sure I see what is this metric

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -130,6 +130,37 @@ message OutputDebugInfo {
...
+  repeated Job running_jobs = 2;
+  repeated ModuleStats modules_stats = 3;
+}
+
+message Job {
+    uint64 stage = 1;
+    repeated string modules = 2;
+    uint64 start_block = 3;
+    uint64 stop_block = 4;
+    uint64 processed_blocks = 5;
```
maoueh: I think `processed_block_count` is better since there clearer that is a literal value and not a list of processed block.

maoueh: Applies to other `..._blocks` and `..._bytes` and the like.

proto/sf/substreams/rpc/v2/service.proto
```diff
@@ -130,6 +130,37 @@ message OutputDebugInfo {
 
 message ModulesProgress {
   repeated ModuleProgress modules = 1;
```
maoueh: I think we shoul deprecate module progress right away

maoueh: I think I would like to see some kind of "completeness" message where a percentage of completion is given. Once live (inside "live" segment), completion would be 100%. That is a key metric that will make implementation of sink must easier, for most cases, a simple `completion` like that reported each Xs is quite enough.

maoueh (changes requested):
I like it!

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: rpc calls is the only implementation of that currently...

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: I was thinking twice per second. No heavy computations there...

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: total could be in seconds. But it would be stored in milliseconds...

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: yes, an error on a module on a tier2, summed up here...
Allows the CLI to report back to the user if a specific module caused errors... Maybe it is not relevant, but since there are retries, it could be interesting information when debugging.

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: I'll add a comment: running a store in chunks:
[0-1000] [1000-2000] [missing chunk...] [3000-4000]
the highest-merged-block for that module is 2000. That information is very relevant if you are trying to figure out why you're not receiving any data from a mapper that depends on this module (imagine that block 2012 is excruciatingly slow...)

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: ok, it makes sense. Not many clients will be parsing it anyway, except the GUI, and we'll change it at the same time. noted.

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: hmmmm not a big fan of oneof... they are harder to work with and don't bring benefits in this case, where we have 9 common fields and 4 fields specific to one of the branch (and they can be 0 without problem!)...

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: I believe it could be globally resolved (from iniital store lookups), but lazily updated (if we are not at HEAD yet, and someone generates more files, we don't get notified)

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: hmmmm... could be complex, these metrics will serve "production" as well as "development" requests.
If you have a start-block at 15M, there is the notion of 'completeness' in back-processing before your start-block (why is it so slow before I get any data ?), then the notion of completeness for substreams on the whole chain... In this case, the "highest_processed_block" per module does give a sense of completeness, the definition of which can be (easily) computed on the client side.

I'm already thinking of cases where it will be difficult to determine "which" completeness means the most to the user (ex: development mode, or a substreams with only a mapper -- which can start at any block, production mode where the mapper cache-files have been produced only for certain specific ranges, ex: [10,000-20,000] but the dependency stores are ready up to block 15M...)

My approach here is to give correctly aggregated data, but I'm trying to avoid "computed subjective data", and I see 'completeness' as one of them...

You are welcome to change my mind :D

proto/sf/substreams/rpc/v2/service.proto
abourget, line 163: We've phased out `Job` everywhere, used `WorkUnit` instead.. perhaps this should look similar. Is it refering to the same work.Unit ?

proto/sf/substreams/rpc/v2/service.proto
abourget: highest isn't so meaningful if there are plenty of segments that aren't finished, no? So perhaps highest contiguous? But then, it doesn't mean you can do anything at that highest block (maybe the first segment isn't finished, and everything on next stages are blocked on it)
Unsure of the use here.

proto/sf/substreams/rpc/v2/service.proto
abourget: I'd go for `ms` too, you have an int here.. so we could never have lower than a second.. which seems poor for a perf metric :)

proto/sf/substreams/rpc/v2/service.proto
abourget: There's a convention for those `_count` suffixes, it's that it's always singular before the `_count`.. so that would be, for both: `total_store_read_count` and `total_error_count` .. let's unify all the rest this way too.

proto/sf/substreams/rpc/v2/service.proto
abourget: Yes that's good. I'd also want to know the time spent in merging, and/or a way to understand we're blocked on merging, and that's what's slowing things down.

Actually, I'd like to know in the ModuleStats whether the module is currently being merged.. so across the updates, you could see if either the blocks are ramping up, or if the "merging: bool" field is often true.

That information is available on a per-stage basis in the Scheduler.

I'm wondering if we shouldn't have a StageStats rather than a Module Stats.. or perhaps those values nested (`StageStats{modules ModuleStats{{}, {}, {}}`), so you'd look down the Module only if you really need to.

 I imagine the `GUI` to display only stages at some point.. and perhaps as you hover, or select the row, you can see which modules are included in that stage, and get a breakdown (for instance, to know if a given module is still merging, and other modules have stopped, and at the stage level, you would be "merging" all the while any of the module below is merging.

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: Ok, I will remove `highest_processed_block` because in the case of the mapper, we don't know and the definition is not obvious (if I'm asking for blocks 1000->2000 and the execout cache already has block 15M to 16M, what will 'highest_processed_block` show ?)

I will only keep the `highest_merged_block` value, for stores.
The `running_jobs` will allow further analysis of current state of output mapper jobs.

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: Our documentation mentions "tier2 jobs" or "running jobs on tier2", and we have flags like `X-Sf-Substreams-Parallel-Jobs`

I think it is easy enough to grasp the meaning of a 'job' in that context. The back-end implementation uses WorkUnit to determine a **segment** and a **stage**, which is the unit that can be made into a job. It is not exactly the same meaning, and "work unit" everywhere would make documentation cumbersome.

proto/sf/substreams/rpc/v2/service.proto
sduchesneau: We don't need "stats" for stages as much as we need to know the completed blockranges (the jobs are per-stage so they complete the information for progress...)
I'll add `stages` to the modules progress that gives you all the completed ranges.
I'll remove the list of modules from the job, because they can be looked up on the stages.
Stats must be per-module because this is how you figure out issues, but the progress must be per stage.

abourget:
Are we, with this PR, moving reporting to stage-based progress? And is it clear which modules are within each stage?

cmd/substreams/info.go
```diff
@@ -9,22 +9,24 @@ import (
...
+	"github.com/streamingfast/substreams/pipeline/outputmodules"
 )
 
 //	var manifestCmd = &cobra.Command{
 //		Use:          "manifest",
 //		SilenceUsage: true,
 //	}
 var infoCmd = &cobra.Command{
-	Use:   "info [<manifest_file>]",
+	Use:   "info [<manifest_file> [<output_module>]]",
```
maoueh, line 20:
```suggestion
	Use:   "info [<manifest_file>] [<output_module>]",
```

docs/release-notes/change-log.md
```diff
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Highlights
+
+* Complete redesign of the progress messages: statistics are now aggregated between tier2 and tier1 and represent a snapshot of the current state.
+* tier1 progress messages now include details of current running parallel stages and a breakdown of where the time is spent in each module.
+
```
maoueh, line 16: We should mention the deprecation of `ModuleProgressMessage` and telling this don't emit anything anymore.

proto/sf/substreams/intern/v2/service.proto
```diff
@@ -19,14 +19,46 @@ message ProcessRangeRequest {
 }
 
 message ProcessRangeResponse {
```
maoueh, line 21: Can we describe the update flow (each 500ms) and state and general contract of the progress message at the RPC level.

Let's add mucho comment in the proto definition please

proto/sf/substreams/rpc/v2/service.proto
maoueh: Still relevant for higher element in the proto, good job on the individual fields!

docs/release-notes/change-log.md
maoueh: We need to also better document what is the impact of users. When we release that what happen for:
- Users that do not upgrade `substreams` yet?

maoueh (approved):
Approving but I left some comments.

PR #280, opened 2023-08-11
Rust tutorial - Using Results

enoldev:


maoueh:
Pretty good, I think we should merge it, mostly as-is and simply improve on later iterations.

enoldev:
Ok! I'll merge this first, and add the other tutorials after

PR #296, opened 2023-09-01
Add validation for start block number vs first streamable block of the chain

### LINKED ISSUE #290: Substreams module's `initialBlock` is not considered as the chain's genesis block
On `arb-one.streamingfast.io:443`, if you try a `substreams.spkg` for which the `initialBlock` is `0` (and you don't specify any range which means full chain), then you are faced with a weird error message from the server:

```
2023-08-30T23:19:36.051-0400 ERRO (sink-postgres) substreams encountered a retryable error {"error": "rpc error: code = Unknown desc = rpc error: code = InvalidArgument desc = start block 22207818 is after stop block 70000"}
```

This happens as the `tier2` makes a Firehose request and the error comes from there. And since we do 10 parallel request or more, the `stop block` varies as it's a race to the first one that succeed in forward *its* error.

There is 2 sides to this issue IMO:
- The `initialBlock: 0` in my opinion should always be considered as the first streamable block of the chain
- Validation is missing when an `initialBlock != 0 && initialBlock < firstStreamableBlock` in which case the error should be clear
    - Side missing, validated that the gRPC error code going out is `InvalidArgument` and not `rpc error: code = Unknown desc = rpc error: code = InvalidArgument` (which is an Unknown error whose message is a stringified gRPC InvalidArgument error so consumer see `Unknown` error but it should have seen `InvalidArgument`)

colindickson:
Added checks in blocks method of Tier1 service to ensure that the requested start block number is equal to or greater than the first streamable block of the chain. This prevents errors caused by invalid block number requests.

### DIFF
--- a/docs/release-notes/change-log.md
+++ b/docs/release-notes/change-log.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Bug fixes
+
+* If the initial block or start block is less than the first block in the chain, the substreams will now start from the
+  first block in the chain. Previously, setting the initial block to a block before the first block in the chain would
+  cause the substreams to hang.
+
 ## v1.1.12
 
 ### Highlights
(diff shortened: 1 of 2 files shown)


maoueh:
Looks good, please update the CHANGELOG (let's add other ones you fixed if they are still missing)

maoueh (approved):
CHANGELOG missing

PR #298, opened 2023-09-05
Docs - Rust tutorial: Option struct

enoldev:


docs/tutorials/rust/option.md
```diff
@@ -0,0 +1,78 @@
...
+
+In a success case, you pass the user identifier as a parameter and the function returns the corresponding username. However, **what happens if the function does not have a username for the given identifier?** A possible solution is to return an empty string:
+- If the function **is able to retrieve the data**, then the string returned is the username.
+- If the function **is NOT able to retrieve the data**, then the string returned is the empty string (`''`).
+
+Although this is a valid approach, it creates hidden logic that is not visible unless to deep dive into the function code.
+
+## The Solution
+
+Rust provides a better way of dealing with these situations by using the `Option<T>` enum. This enum has two possible values: `Some<T>` (used when the returned value is present) and `None` (used when the returned value is not present). Therefore, the previous function can be refactored:
```
maoueh:
```suggestion
Rust provides a better way of dealing with these situations by using the `Option<T>` enum. This enum has two possible values: `Some(T)` (used when the returned value is present) and `None` (used when the returned value is not present). Therefore, the previous function can be refactored:
```

docs/tutorials/rust/option.md
```diff
@@ -0,0 +1,78 @@
...
+
+```rust
+fn get_username_by_id(id: u32) -> Option<String> {
+    // function body
+}
+```
+
+Now, the function works as follows:
+- If the function **is able to retrieve the data**, then a `Some` struct containing the string is returned.
```
maoueh:
```suggestion
- If the function **is able to retrieve the data**, then a `Some` value containing the string is returned.
```

docs/tutorials/rust/option.md
```diff
@@ -0,0 +1,78 @@
...
+- If the function **is NOT able to retrieve the data**, then a `None` struct is returned.
```
maoueh:
```suggestion
- If the function **is NOT able to retrieve the data**, then a `None` value is returned.
```

docs/tutorials/rust/option.md
```diff
@@ -0,0 +1,78 @@
...
+}
+```
+1. Given a user identifier, return the corresponding username if exists.
+2. If `id == 1`, then a `Some` struct containing the string is returned.
+3. If `id == 2`, then a `Some` struct containing the string is returned.
+4. If `id` does not match with any of the provided integers, then a `None` struct is returned.
+
+## Using Options
+
+The `Option<T>` struct contains two helper methods to check if the returned type is `Some` or `None`: the `.is_some()` and `.is_none()` methods. Let's see how to use these methods:
```
maoueh, line 51: There should be a note saying that using pattern matching is the preferred approach because it forces to cover all the enum's values.

maoueh (approved):
Small stuff, approving nonetheless

PR #311, opened 2023-09-18
Feature/deployable

sduchesneau:
### Added

* Sink configs can now use protobuf annotations (aka Field Options) to determine how the field will be interpreted in
  substreams.yaml:
  * `load_from_file` will put the content of the file directly in the field (string and bytes contents are supported).
  * `zip_from_folder` will create a zip archive and put its content in the field (field type must be bytes).

  Example: 
  ```
  import "sf/substreams/v1/options.proto";
  optional bytes schema = 1 [ (sf.substreams.v1.options).load_from_file = true ];`
  optional bytes extra_config_files = 2 [ (sf.substreams.v1.options).zip_from_folder = true ];`
  ```
* `substreams info` command now properly displays the sink configs, optionally writing the bytes fields to disk with `--dump-sinkconfig-bytes-to-disk` flag

### Removed

*  The `@path/to/textfile` and `@@path/to/binaryfile` notations under sink_configs in substreams.yaml are now DEPRECATED (they will still work but with a warning).


docs/release-notes/change-log.md
```diff
@@ -4,6 +4,27 @@ All notable changes to this project will be documented in this file.
...
+* Sink configs can now use protobuf annotations (aka Field Options) to determine how the field will be interpreted in
+  substreams.yaml:
+  * `load_from_file` will put the content of the file directly in the field (string and bytes contents are supported).
+  * `zip_from_folder` will create a zip archive and put its content in the field (field type must be bytes).
+
+  Example: 
+  ```
+  import "sf/substreams/v1/options.proto";
+  optional bytes schema = 1 [ (sf.substreams.v1.options).load_from_file = true ];`
+  optional bytes extra_config_files = 2 [ (sf.substreams.v1.options).zip_from_folder = true ];`
```
maoueh: Let's showcase the `message ...` part as it's not clear like that where those are used. Would probably use a "complete" message.

manifest/sink.go
```diff
@@ -35,49 +52,168 @@ func (r *Reader) loadSinkConfig(pkg *pbsubstreams.Package, m *Manifest) error {
...
+func fieldResolver(msgDesc *desc.MessageDescriptor) func(string) (opts *pbsubstreams.FieldOptions, isBytes bool) {
+	return func(name string) (opts *pbsubstreams.FieldOptions, isBytes bool) {
+		options := &pbsubstreams.FieldOptions{}
+		fqdn := msgDesc.GetFullyQualifiedName() + "." + name
+		for _, fd := range msgDesc.GetFields() {
+			if fd.GetFullyQualifiedJSONName() == fqdn {
+				isBytes := fd.GetType() == descriptorpb.FieldDescriptorProto_TYPE_BYTES
+				if opts := fd.GetFieldOptions(); opts != nil {
+					opts.ProtoReflect().Range(func(desc protoreflect.FieldDescriptor, val protoreflect.Value) bool {
+						if desc.TextName() != "[sf.substreams.v1.options]" {
```
maoueh: The literal is repeated, let's extract it (can we even extract from the Message directly).

manifest/sink.go
```diff
@@ -117,3 +294,33 @@ func convertYAMLtoJSONCompat(i any, resolvePath func(in string) string) (out any
 	}
 	return i, nil
 }
+
+func addFiles(w *zip.Writer, basePath string) error {
+	return filepath.Walk(basePath, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		path = strings.Replace(strings.TrimPrefix(strings.TrimPrefix(path, basePath), string(filepath.Separator)), "\\", "/", -1)
```
maoueh: Really hard to read and reason about

maoueh: At the very minimum, a quick comment is required.

proto/sf/substreams/v1/options.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+package sf.substreams.v1;
+
+option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/v1;pbsubstreams";
+import "google/protobuf/descriptor.proto";
+
+message FieldOptions {
+  // this flag informs the `substreams pack` command that it should treat the corresponding manifest value as a path to a file, putting its content as bytes in this field.
```
maoueh:
```suggestion
  // This option informs the `substreams pack` command that it should treat the corresponding manifest value as a path to a file, putting its content as bytes in this field. 
```

proto/sf/substreams/v1/options.proto
```diff
@@ -0,0 +1,19 @@
...
+
+option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/v1;pbsubstreams";
+import "google/protobuf/descriptor.proto";
+
+message FieldOptions {
+  // this flag informs the `substreams pack` command that it should treat the corresponding manifest value as a path to a file, putting its content as bytes in this field. 
+  // must be applied to a `bytes` or `string` field
+  bool load_from_file = 1;
+
+  // this flag informs the `substreams pack` command that it should treat the corresponding manifest value as a path to a folder, zipping its content and putting the zip content as bytes in this field.
```
maoueh:
```suggestion
  // This option informs the `substreams pack` command that it should treat the corresponding manifest value as a path to a folder, zipping its content and putting the zip content as bytes in this field.
```

proto/sf/substreams/v1/options.proto
```diff
@@ -0,0 +1,19 @@
...
+  // must be applied to a `bytes` field
+  bool zip_from_folder = 2;
+}
+
+extend google.protobuf.FieldOptions {
+  optional FieldOptions options = 2200;
```
maoueh: I think there is a central repo that list used/unsued value here, did you see something about that?

manifest/sink.go
```diff
@@ -35,49 +52,168 @@ func (r *Reader) loadSinkConfig(pkg *pbsubstreams.Package, m *Manifest) error {
...
+	}
+	pkg.SinkConfig = &anypb.Any{
+		TypeUrl: m.Sink.Type,
+		Value:   pbBytes,
+	}
+
+	return nil
+}
+
+func SinkConfigAsString(pkg *pbsubstreams.Package, dumpBytesToDisk bool) (string, error) {
```
maoueh: I dislike the fact that converting to string also dumps to file. In fact more reading lead me to think that this function is weirdly named.

It also deals with representational aspect, the output being a new-line separated string. The function should return a `[]string` and let the caller decide what to do.

I also think the path where to write things should be pass in by the caller. Inferring `tmp` seems unecessary, the caller should deal with this decision.

maoueh (changes requested):

Issue #309, opened 2023-09-18
Network config override for Substreams manifests


jubeless:
```
SUBSTREAMS_ENDPOINTS_CONFIG_POLYGON=polygonl.friend.com:443
SUBSTREAMS_ENDPOINTS_CONFIG_MAINNET=ethmainn.friend.com:443
```
To support:
`substreams run myfile.polygon.yaml`
and:
`substreams run myfile.yaml`  // where network is `mainnet`
`myfile.polygon.yaml`:
```yaml
network: polygon
#network: mumbai
#network: goerli
#network: goerli
#network: eip1544
```
Align with: https://thegraph.com/docs/en/developing/supported-networks/
1. SUBSTREAMS_ENDPOINTS_CONFIG_\[network\] and defined in the manifest. (this is what needs implementing)
2. SUBSTREAMS_ENDPOINT viper-based config
3. -e value
4. default `-e` value (remove the default value, issue a deprecation notice)
- Error: missing endpoint, specify -e or `SUBSTREAMS_ENDPOINTS_CONFIG_[network]` as defined in `network` field of the Substreams manifest.

sduchesneau:
* Substreams will have a hardcoded map of 'network -> endpoint' to use if the environment variables are not used. It will point to all streamingfast's endpoints, or Pinax for the ones that we don't support but they do (ex: antelope)
* This hardcoded map will come with a "TODO" comment saying that this is temporary until we have something "distributed" (like, the gateway advertising available endpoints or whatever)
* If a substreams manifest does not specify the 'Network', the default is to consider it to be `mainnet`.
* the `-e` flag always overrides this whole mechanism

The idea is also that the 'subservicesd' component, where we deploy, will use those values...

PR #316, opened 2023-09-26
Feature/network config override

colindickson:
solves https://github.com/streamingfast/substreams/issues/309

### DIFF
--- a/cmd/substreams/gui.go
+++ b/cmd/substreams/gui.go
@@ -88,13 +88,9 @@ func runGui(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("read manifest %q: %w", manifestPath, err)
 	}
 
-	endpoint := mustGetString(cmd, "substreams-endpoint")
-	envEndpoint := manifest.GetNetworkEndpointFromEnvironment(pkg.Network)
-	if endpoint == "" && envEndpoint != "" {
-		endpoint = envEndpoint
-	}
-	if endpoint == "" {
-		return fmt.Errorf("missing endpoint, specify -e or SUBSTREAMS_ENDPOINTS_CONFIG_[network] as defined in network field of the Substreams manifest")
+	endpoint, err := manifest.ExtractNetworkEndpoint(pkg.Network, mustGetString(cmd, "substreams-endpoint"))
+	if err != nil {
+		return fmt.Errorf("extracting endpoint: %w", err)
 	}
 
 	substreamsClientConfig := client.NewSubstreamsClientConfig(
--- a/cmd/substreams/gui.go
+++ b/cmd/substreams/gui.go
@@ -19,7 +19,7 @@ import (
 
 func init() {
 	guiCmd.Flags().String("substreams-api-token-envvar", "SUBSTREAMS_API_TOKEN", "name of variable containing Substreams Authentication token")
-	guiCmd.Flags().StringP("substreams-endpoint", "e", "mainnet.eth.streamingfast.io:443", "Substreams gRPC endpoint")
+	guiCmd.Flags().StringP("substreams-endpoint", "e", "mainnet.eth.streamingfast.io:443", "Substreams gRPC endpoint. If empty, will be replaced by the SUBSTREAMS_ENDPOINT_{network_name} environment variable, where `network_name` is determined from the substreams manifest")
 	guiCmd.Flags().Bool("insecure", false, "Skip certificate validation on GRPC connection")
 	guiCmd.Flags().Bool("plaintext", false, "Establish GRPC connection in plaintext")
 	guiCmd.Flags().StringSliceP("header", "H", nil, "Additional headers to be sent in the substreams request")
--- a/cmd/substreams/gui.go
+++ b/cmd/substreams/gui.go
@@ -79,13 +79,6 @@ func runGui(cmd *cobra.Command, args []string) error {
 
 	outputModule := args[0]
 
-	substreamsClientConfig := client.NewSubstreamsClientConfig(
-		mustGetString(cmd, "substreams-endpoint"),
-		tools.ReadAPIToken(cmd, "substreams-api-token-envvar"),
-		mustGetBool(cmd, "insecure"),
-		mustGetBool(cmd, "plaintext"),
-	)
-
 	manifestReader, err := manifest.NewReader(manifestPath)
 	if err != nil {
 		return fmt.Errorf("manifest reader: %w", err)
@@ -94,6 +87,23 @@ func runGui(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("read manifest %q: %w", manifestPath, err)
 	}
+
+	endpoint := mustGetString(cmd, "substreams-endpoint")
+	envEndpoint := getNetworkEndpointFromEnvironment(pkg.Network)
+	if endpoint == "" && envEndpoint != "" {
+		endpoint = envEndpoint
+	}
+	if endpoint == "" {
+		return fmt.Errorf("missing endpoint, specify -e or SUBSTREAMS_ENDPOINTS_CONFIG_[network] as defined in network field of the Substreams manifest")
+	}
+
+	substreamsClientConfig := client.NewSubstreamsClientConfig(
+		endpoint,
+		tools.ReadAPIToken(cmd, "substreams-api-token-envvar"),
+		mustGetBool(cmd, "insecure"),
+		mustGetBool(cmd, "plaintext"),
+	)
+
 	params := mustGetStringArray(cmd, "params")
 	if err := manifest.ApplyParams(params, pkg); err != nil {
 		return err
--- a/cmd/substreams/gui.go
+++ b/cmd/substreams/gui.go
@@ -3,11 +3,6 @@ package main
 import (
 	"errors"
 	"fmt"
-	"os"
-	"path/filepath"
-	"strings"
-	"time"
-
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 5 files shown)


cmd/substreams/run.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 func init() {
-	runCmd.Flags().StringP("substreams-endpoint", "e", "mainnet.eth.streamingfast.io:443", "Substreams gRPC endpoint")
+	runCmd.Flags().StringP("substreams-endpoint", "e", "", "Substreams gRPC endpoint")
```
maoueh: We must not make this breaking change. It will be breaking for those not defining a `network` in their manifest.

- If `network` is defined, we should offer a default value if none is provided, based on the `network`.
- If the `network` is undefined, we should fallback to `mainnet.eth.streamingfast.io` and issue a deprecation warning with a clear message and upgrade path.

The changelog show also clearly define the change and upgrade path.

cmd/substreams/run.go
```diff
@@ -78,6 +78,15 @@ func runRun(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("read manifest %q: %w", manifestPath, err)
 	}
 
+	endpoint := mustGetString(cmd, "substreams-endpoint")
+	envEndpoint := manifest.GetNetworkEndpointFromEnvironment(pkg.Network)
+	if endpoint == "" && envEndpoint != "" {
+		endpoint = envEndpoint
+	}
+	if endpoint == "" {
+		return fmt.Errorf("missing endpoint, specify -e or SUBSTREAMS_ENDPOINTS_CONFIG_[network] as defined in network field of the Substreams manifest")
```
maoueh:
```suggestion
		return fmt.Errorf("missing endpoint, specify -e or SUBSTREAMS_ENDPOINTS_CONFIG_[network] as defined in 'network' field of the Substreams manifest")
```

cmd/substreams/run.go
maoueh: The documentation of the flag should be expanded to document the new behavior more.

maoueh (changes requested):

cmd/substreams/run.go
sduchesneau: Why not make this a breaking change ? 
It will break only on CLI and the fix can be pretty obvious, ex:
`Error: cannot determine network from the manifest and no endpoint specified. Try "-e mainnet.eth.streamingfast.io:443"`

sduchesneau: I do agree with having our own endpoints as default for specific networks if there is no `-e` override. (ex: `matic -> polygon.streamingfast.io:443`) I don't know who was against that in the start ??

manifest/utils.go
```diff
@@ -27,3 +29,22 @@ func (s *mapSlice) UnmarshalYAML(n *yaml.Node) error {
...
 	return nil
 }
+
+func GetNetworkEndpointFromEnvironment(networkName string) string {
+	networkName = mapNetworkNameToCanonicalName(networkName)
+	networkEndpoint := os.Getenv(fmt.Sprintf("SUBSTREAMS_ENDPOINTS_CONFIG_%s", strings.ToUpper(networkName)))
+	return networkEndpoint
+}
+
+func mapNetworkNameToCanonicalName(networkName string) string {
```
sduchesneau: After thinking about this, it may be a bad idea: the manifest would say something "wrong" like "polygon", and the environment variable SUBSTREAMS_ENDPOINTS_CONFIG_POLYGON would not work... this sucks... Better keep it unchanged...

cmd/substreams/run.go
colindickson: isn't the network a required field in the manifest? it should always be defined.  its the endpoint that might not be determined from the network value.  i don't mind doing a print of a deprecation warning (if the network is `mainnet`), but i would prefer a breaking change.

sduchesneau: Requiring that the network field is populated is harder --> it is a breaking change that prevents people from syncing any spkg created before that change. we must allow an override of the endpoint when the network is not set.

maoueh: `network` is not mandatory. I think we can avoid the breaking change that's why I'm proposing it. Lot's of people are doing today `substreams run <manifest>` with a Substreams manifest that contains no `network`, why breaking them when we can offer a nicer upgrade path.

If everyone thinks breaking is better, I won't resist.

sduchesneau (dismissed):
~LGTM~

Actually wait, @colindickson please revisit the deprecation warning as mentionned below!

cmd/substreams/run.go
```diff
@@ -78,6 +79,16 @@ func runRun(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("read manifest %q: %w", manifestPath, err)
 	}
 
+	endpoint, err := manifest.ExtractNetworkEndpoint(pkg.Network, mustGetString(cmd, "substreams-endpoint"))
+	if err != nil {
+		fmt.Println(networkDefaultEndpointDeprecationWarning)
```
sduchesneau: This deprecation warning should live outside of ExtractNetworkEndpoint:
```
substreamsEndpointFlag := mustGetString(cmd, "substreams-endpoint")
if pkg.Network == "" && substreamsEndpointFlag == "" {
 // trigger deprecation warning logic
} else {
// manifest.ExtractNetworkEndpoint(...)
}
```

sduchesneau: This way the error is returned for every other case

sduchesneau: Also, there should be a CHANGELOG entry and README entry about this:
"here's how you can easily handle different networks in substreams: set your endpoints to each environment variable, ....." etc.

cmd/substreams/gui.go
```diff
@@ -19,7 +20,7 @@ import (
 
 func init() {
 	guiCmd.Flags().String("substreams-api-token-envvar", "SUBSTREAMS_API_TOKEN", "name of variable containing Substreams Authentication token")
-	guiCmd.Flags().StringP("substreams-endpoint", "e", "mainnet.eth.streamingfast.io:443", "Substreams gRPC endpoint")
+	guiCmd.Flags().StringP("substreams-endpoint", "e", "", "Substreams gRPC endpoint. If empty, will be replaced by the SUBSTREAMS_ENDPOINT_{network_name} environment variable, where `network_name` is determined from the substreams manifest")
```
maoueh: We are missing the part for the default based on `network`.

manifest/utils.go
```diff
@@ -27,3 +29,23 @@ func (s *mapSlice) UnmarshalYAML(n *yaml.Node) error {
 
 	return nil
 }
+
+func ExtractNetworkEndpoint(networkFromManifest, fromFlag string) (string, error) {
+	if fromFlag != "" {
+		return fromFlag, nil
+	}
+	if networkFromManifest == "" {
+		return "", fmt.Errorf("cannot determine endpoint. Either specify it with a flag, `-e mainnet.eth.streamingfast.io:443` or use the 'Network' field in the manifest, matching with SUBSTREAMS_ENDPOINTS_CONFIG_[network] environment variable")
```
maoueh:
```suggestion
		return "", fmt.Errorf("cannot determine endpoint: either specify it with a flag, `-e mainnet.eth.streamingfast.io:443` or use the 'network' field in the manifest, matching with SUBSTREAMS_ENDPOINTS_CONFIG_[network] environment variable")
```

manifest/utils.go
```diff
@@ -27,3 +29,23 @@ func (s *mapSlice) UnmarshalYAML(n *yaml.Node) error {
...
+	}
+
+	endpoint := GetNetworkEndpointFromEnvironment(networkFromManifest)
+	if endpoint == "" {
+		return "", fmt.Errorf("cannot determine endpoint for network %q. Make sure that you set SUBSTREAMS_ENDPOINTS_CONFIG_%s environment variable to a valid endpoint", networkFromManifest, strings.ToUpper(networkFromManifest))
```
maoueh:
```suggestion
		return "", fmt.Errorf("cannot determine endpoint for network %q: make sure that you set SUBSTREAMS_ENDPOINTS_CONFIG_%s environment variable to a valid endpoint", networkFromManifest, strings.ToUpper(networkFromManifest))
```

cmd/substreams/utils.go
```diff
@@ -54,3 +54,5 @@ func readStopBlockFlag(cmd *cobra.Command, startBlock int64, flagName string, wi
 
 	return endBlock, nil
 }
+
+var networkDefaultEndpointDeprecationWarning string = "DEPRECATION WARNING: default endpoint is deprecated. Please specify a network endpoint using the --substreams-endpoint flag (-e) or by specifying an environment variable in the form SUBSTREAMS_ENDPOINTS_CONFIG_{network} where 'network' is the name of the network for this Substreams"
```
maoueh:
```suggestion
var networkDefaultEndpointDeprecationWarning string = "DEPRECATION WARNING: Default endpoint is deprecated. Define a 'network' field in your Substreams manifest in which case the endpoint will be inferred for you. You can also manually specify the endpoint to use by using '--substreams-endpoint <endpoint>' flag ('-e <endpoint>') or by specifying an environment variable in the form SUBSTREAMS_ENDPOINTS_CONFIG_{network} where 'network' is the name of the network defined in your Substreams manifest"
```

maoueh: We should also think about linking the list of network somewhere (@enoldev for info about the upcoming feature).

maoueh (approved):
Some small wording suggestions, feel free to adjust/pick/skip

PR #330, opened 2023-10-26
store-free output module optimization

colindickson:
optimized start of output stream in developer mode when start block is in reversible segment and output module does not have any stores in its dependencies

### DIFF
--- a/pipeline/resolve.go
+++ b/pipeline/resolve.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/streamingfast/substreams/manifest"
 	"sync/atomic"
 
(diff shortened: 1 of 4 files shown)


pipeline/resolve.go
```diff
@@ -94,6 +106,10 @@ func computeLinearHandoffBlockNum(productionMode bool, startBlock, stopBlock uin
 	if err != nil {
 		return startBlock, nil
 	}
+
+	if !stateRequired {
```
sduchesneau: move this to line 105 to avoid computation

sduchesneau (changes requested):
LGTM, just skip the computation of maxHandoff

PR #338, opened 2023-11-08
Made fixes for Windows when using init

Dirt-Nasty:


cmd/substreams/init.go
```diff
@@ -210,7 +210,7 @@ func renderProjectFilesIn(project templates.Project, absoluteProjectDir string)
 	}
 
 	for relativeFile, content := range files {
-		file := path.Join(absoluteProjectDir, strings.ReplaceAll(relativeFile, "/", string(os.PathSeparator)))
+		file := path.Join(absoluteProjectDir, relativeFile)
```
maoueh: I'm surprised of this change, here we use `os.WriteFile` so I'm expecting that `file` should use OS path separator. So I would see the change more as:

```suggestion
		file := filepath.Join(absoluteProjectDir, strings.ReplaceAll(relativeFile, "/", string(os.PathSeparator)))
```

codegen/templates/ethereum_project.go
```diff
@@ -138,10 +138,9 @@ func (p *EthereumProject) Render() (map[string][]byte, error) {
 			ethereumProjectEntry = "src/multiple_contracts_lib.rs.gotmpl"
 		}
 
-		content, err := ethereumProject.ReadFile(filepath.Join("ethereum", ethereumProjectEntry))
+		content, err := ethereumProject.ReadFile(path.Join("ethereum", ethereumProjectEntry))
```
maoueh: That's a good catch, the usage of `filepath` was wrong because `ethereumProject` is a bit special and path in there is always using `/`. 

Let's however avoid usage of `path.Join` and let's add a comment to explain why it's hard-coded:
```suggestion
        // We use directly "/" here as `ethereumProject` is an embed FS and always uses "/"
		content, err := ethereumProject.ReadFile("ethereum" + "/" + ethereumProjectEntry)
```

maoueh (changes requested):

cmd/substreams/init.go
maoueh: Ok `absoluteProjectDir` is also problematic since it's wrongly built, need to change this line [absoluteProjectDir](https://github.com/streamingfast/substreams/blob/b9cef0fef35e45c5f3cd0f7e976915f09aad6ded/cmd/substreams/init.go#L75) to become `	absoluteProjectDir := filepath.Join(absoluteWorkingDir, projectName)
`

Dirt-Nasty: After the suggested change [here](https://github.com/streamingfast/substreams/blob/b9cef0fef35e45c5f3cd0f7e976915f09aad6ded/cmd/substreams/init.go#L75) init is failing

`Error: render Ethereum Ethereum Mainnet project: removing directory .: RemoveAll .: invalid argument and write file: open F:\Projects\test_init\new_code\substreams.yaml: The system cannot find the path specified.`

Dirt-Nasty: Ahhh [this](https://github.com/streamingfast/substreams/blob/b9cef0fef35e45c5f3cd0f7e976915f09aad6ded/cmd/substreams/init.go#L186) need to change as well to use filepath and not path.

maoueh (approved):

PR #354, opened 2023-12-14
Pass substreams endpoint to `service serve`

### LINKED ISSUE #353: `service serve` and non-mainnet substreams
When I run `substreams alpha service serve -e <endpoint>` I expect the sink to use that endpoint when consuming substreams module. But it doesn't seem to be the case and it uses the default mainnet endpoint.

Looks like it's not passing that argument to the sink docker manifest.

YaroShkvorets:
Fixes #353

Allows to use `--endpoint` flag to pass substreams endpoint to connect to.


sduchesneau:
lgtm!
If you specify another network in your substreams.yaml (ex: `network: polygon`) it should point to the default (streamingfast...) endpoint.

Since this is a dev tool, it should be very flexible, so I'm good with this!

PR #360, opened 2023-12-24
Updated sink-* to service *

hashirpm:
From Ethereum address to SQL tutorial still has `substreams alpha sink-serve` and `substreams alpha sink-deploy` which is updated to `substreams alpha service serve` and `substreams alpha service deploy`


enoldev:
We'll get rid of this tutorial and include a version in the new docs. Thanks for flagging.

PR #375, opened 2024-01-15
--init-config flag in substreams init

Dirt-Nasty:
`substreams init --init-config`

I've added an --init-config flag for use in substreams init.

This allows you to use a local config file rather than the command prompts to spin up a substreams projects.

The config should be in the root directory of wherever you're running the substreams init command from.

```
{
    "project_name" : "testing_local",
    "protocol" : "Ethereum",
    "chain" : "Mainnet",
    "contracts" : [
        {
            "address" : "0xB4Ac9F231D2469769A0a56eee5559Fd4613ed3fB",
            "short_name" : "loans_interface",
            "start_block" : "10600000",
            "abi_path" : "./abis/LoansInterface.json"
        }
    ]
}
```
The main use case here is to use local abis for non verified contracts rather than relying on the block explorer.

`contracts` can be an array to support multiple contracts.
`start_block` is currently not used as the explorer is able to handle to retrieving the start block without contract verification.
`abi_path` should point to your local abi file.

All fields should be present in the config.

I've tested this on the following scenarios:
- Single & Multiple substreams init with the init-config flag
- Single & Multiple substreams init without the init-config flag
- Proxy Contracts without the init-config flag

Some things to note:
- Its probably more preferable to use a yaml than json
- Rather than a bool flag allow a string to the location of the config file
- This needs better error handling I'm sure
- Allow dynamic use where some contracts are available on the explorer and some aren't

### DIFF
--- a/cmd/substreams/init.go
+++ b/cmd/substreams/init.go
@@ -54,15 +54,31 @@ var initCmd = &cobra.Command{
 
 var etherscanAPIKey = "YourApiKeyToken"
 
+type ConfigContract struct {
+	Address    string `json:"address"`
+	ShortName  string `json:"short_name"`
+	StartBlock string `json:"start_block"`
+	ABIPath    string `json:"abi_path"`
+}
+
+type InitConfig struct {
+	ProjectName string           `json:"project_name"`
+	Protocol    string           `json:"protocol"`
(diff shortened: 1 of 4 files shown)


maoueh:
It's now possible to save the generator state for easier re-start/re-use.

PR #391, opened 2024-02-06
Add Go sink to the docs

enoldev:


docs/new/consume/stream/go.md
```diff
@@ -0,0 +1,138 @@
...
+            tracer,
+            sink.WithBlockRange(blockRange), // 3.
+        )
+    
+    // ...code omitted...
+}
+```
+1. Run the Substreams in _production_ mode.
+2. Create a Substreams client, which includes the endpoint and the authentication token.
+This will also work if you have a system environment variable, `SUBSTREAMS_API_TOKEN`, with a valid Substreams token.
```
maoueh, line 113: Should we mention that `false, false` parameters represent `insecure, plainText` parameters

docs/new/consume/stream/go.md
```diff
@@ -0,0 +1,138 @@
...
+            ":", // 3.
+            zlog,
+        )
+    
+    // ...code omitted...
+}
+```
+1. Equivalent to the `network` parameter of the CLI. Specify the network to use for params and initialBlocks, overriding the `network`` field in the Substreams package.
+2. The output type emitted by the module.
+3. The range block to execute. `:` means from the start of the chain up to the latest block.
```
maoueh, line 86: We should mention that is a stream forever tracking head of chain.

maoueh (approved):
Small things, approving

PR #427, opened 2024-03-08
Feature/doc readme

ArnaudBger:


manifest/package.go
```diff
@@ -119,12 +119,41 @@ func (r *manifestConverter) manifestToPkg(manif *Manifest) (*pbsubstreams.Packag
 	return pkg, protoDefinitions, r.sinkConfigDynamicMessage, nil
 }
 
+func (m *Manifest) readFileFromName(filename string) ([]byte, error) {
+	if _, err := os.Stat(filepath.Join(m.Workdir, filename)); err == nil {
```
abourget, line 123: this won't be case insensitive

manifest/package.go
```diff
@@ -119,12 +119,41 @@ func (r *manifestConverter) manifestToPkg(manif *Manifest) (*pbsubstreams.Packag
...
+		}
+		return content, nil
+	}
+	return nil, nil
+}
+
 func (r *manifestConverter) convertToPkg(m *Manifest) (pkg *pbsubstreams.Package, err error) {
+	doc := m.Package.Doc
+	if doc == "" {
+		readmeContent, err := m.readFileFromName("README.md")
```
abourget, line 137: you'll have a NotFound error here, you need to check fall back from there for checking README..

manifest/package.go
```diff
@@ -119,12 +119,41 @@ func (r *manifestConverter) manifestToPkg(manif *Manifest) (*pbsubstreams.Packag
...
+		if err != nil {
+			return nil, fmt.Errorf("reading file: %w", err)
+		}
+		if readmeContent == nil {
```
abourget, line 141: if err == nil, you'll never have a `nil` readmeContent. Not useful to check.

PR #430, opened 2024-03-11
Added metrics to see the number of active and total counter of tier2 worker request for graphing purposes

maoueh:
Also fixed readiness label which was labelled as `firehose`.

### DIFF
--- a/app/tier1.go
+++ b/app/tier1.go
@@ -186,7 +186,7 @@ func (a *Tier1App) Run() error {
 			a.logger.Info("waiting until hub is real-time synced")
 			select {
 			case <-forkableHub.Ready:
-				metrics.AppReadiness.SetReady()
+				metrics.AppReadinessTier1.SetReady()
 			case <-a.Terminating():
 				return
 			}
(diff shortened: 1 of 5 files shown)


sduchesneau:
The tier2 is missing a similar metric, and with future naming, I see some confusion...
could you take the time to:

1)  add the same metric on a given tier2 (on the ProcessRange method), named `substreams_tier2_active_requests` and `substreams_tier2_request_count`

2) rename your metrics here from:
* `substreams_active_worker_request` -> `substreams_tier1_active_worker_requests`
* `substreams_worker_request_counter` -> `substreams_tier1_worker_request_count`

(or something simliar) 

?

maoueh:
I had `tier2` in the name initially but I found it confusing too, I tried variation, `tier1` like you suggested was one of them. It wasn't that clear neither, only reading the description felt it helped. It also brought the question that no metrics are currently prefixed. Let's discuss over a call later to decide on the final naming.

Also, we do have already `substreams_active_request`, so should it be used as a "common" for both? On this aspect, let's start with just dealing with what I added and let's add support for more metrics on another PR(s).

Also @matthewdarwin if you have an opinion here.

sduchesneau:
Perfect now!

sduchesneau (approved):
lgtm

PR #432, opened 2024-03-13
Implement use module feature 

ArnaudBger:
Enabling any modules to use another one, with the `use` attribute, and to override its inputs/blockinitial.

### DIFF
--- a/manifest/package.go
+++ b/manifest/package.go
@@ -153,8 +153,7 @@ func (r *manifestConverter) manifestToPkg(manif *Manifest) (*pbsubstreams.Packag
 		return nil, nil, nil, fmt.Errorf("error parsing sink configuration: %w", err)
 	}
 
-	err = handleUseModules(pkg, manif)
-	if err != nil {
+	if err := handleUseModules(pkg, manif); err != nil {
 		return nil, nil, nil, fmt.Errorf("error handling use modules: %w", err)
 	}
 
--- a/manifest/package.go
+++ b/manifest/package.go
@@ -78,7 +78,7 @@ func (r *manifestConverter) validateManifest(manif *Manifest) error {
 			}
 		case "":
 			if s.Use == "" {
-				return fmt.Errorf("stream %q: missing 'use' attribute for a module not specifying any kind", s.Name)
+				return fmt.Errorf("module kind not specified for %q", s.Name)
 			}
 
 			if err := validateModuleWithUse(s); err != nil {
@@ -88,6 +88,7 @@ func (r *manifestConverter) validateManifest(manif *Manifest) error {
 		default:
 			return fmt.Errorf("stream %q: invalid kind %q", s.Name, s.Kind)
 		}
+
 		for idx, input := range s.Inputs {
 			if err := input.parse(); err != nil {
 				return fmt.Errorf("module %q: invalid input [%d]: %w", s.Name, idx, err)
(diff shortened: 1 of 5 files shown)


manifest/package.go
```diff
@@ -76,6 +76,14 @@ func (r *manifestConverter) validateManifest(manif *Manifest) error {
 			if err := validateStoreBuilder(s); err != nil {
 				return fmt.Errorf("stream %q: %w", s.Name, err)
 			}
+		case "":
+			if s.Use == "" {
+				return fmt.Errorf("stream %q: missing 'use' attribute for a module not specifying any kind", s.Name)
```
sduchesneau: the error is not related to "use", it is related to missing module kind.

Error should read `("module kind not specified for %q", s.Name)`. don't confuse people with the use feature...

manifest/package.go
```diff
@@ -76,6 +76,14 @@ func (r *manifestConverter) validateManifest(manif *Manifest) error {
...
+			}
+
+			if err := validateModuleWithUse(s); err != nil {
```
sduchesneau, line 88: This should be called independently of "s.Kind" Move it to line 66.

manifest/package.go
```diff
@@ -90,6 +98,87 @@ func (r *manifestConverter) validateManifest(manif *Manifest) error {
...
 }
 
+func handleUseModules(pkg *pbsubstreams.Package, manif *Manifest) error {
+	packageModulesMapping := make(map[string]*pbsubstreams.Module)
+	for _, module := range pkg.Modules.Modules {
+		packageModulesMapping[module.Name] = module
+	}
+
+	for _, manifestModule := range manif.Modules {
+		if manifestModule.Use != "" {
```
sduchesneau: In Go, try to avoid deeply nested conditions by exiting early:
```
if manifestModule.Use == "" {
continue
}
...
```

manifest/package.go
```diff
@@ -90,6 +98,87 @@ func (r *manifestConverter) validateManifest(manif *Manifest) error {
...
+			moduleWithUse := packageModulesMapping[manifestModule.Name]
+
+			if err := checkEqualInputs(moduleWithUse, usedModule, manifestModule, packageModulesMapping); err != nil {
+				return fmt.Errorf("checking inputs for module %q: %w", manifestModule.Name, err)
+			}
+
+			moduleWithUse.BinaryIndex = usedModule.BinaryIndex
+			moduleWithUse.BinaryEntrypoint = usedModule.BinaryEntrypoint
+
+			if moduleWithUse.InitialBlock == 0 {
```
sduchesneau: Remove these 3 lines completely: since we cannot differentiate between "0" and "nil", let's force the user to always set an initialBlock, so we don't override with the usedModule.

manifest/package.go
```diff
@@ -90,6 +98,87 @@ func (r *manifestConverter) validateManifest(manif *Manifest) error {
...
+		usedModuleInput := usedModule.Inputs[index]
+		if input.GetParams() != nil {
+			if usedModuleInput.GetParams() == nil {
+				return fmt.Errorf("module %q: input %q is not a params type", manifestModuleWithUse.Name, input.String())
+			}
+			if input.GetParams().Value != usedModuleInput.GetParams().Value {
+				return fmt.Errorf("module %q: input %q has different value than the used module %q: input %q", manifestModuleWithUse.Name, input.String(), manifestModuleWithUse.Use, usedModuleInput.String())
+			}
+		}
+		if input.GetSource() != nil {
```
sduchesneau: should be a series of `else if`, because an input is either one of those, never more than one type.

manifest/package.go
```diff
@@ -90,6 +98,87 @@ func (r *manifestConverter) validateManifest(manif *Manifest) error {
...
+			if usedModuleInput.GetStore() == nil {
+				return fmt.Errorf("module %q: input %q is not a store type", manifestModuleWithUse.Name, input.String())
+			}
+			if input.GetStore().GetMode() != usedModuleInput.GetStore().GetMode() {
+				return fmt.Errorf("module %q: input %q has different mode than the used module %q: input %q", manifestModuleWithUse.Name, input.String(), manifestModuleWithUse.Use, usedModuleInput.String())
+			}
+
+			inputStoreModuleName := input.GetStore().ModuleName
+			usedModuleStoreMapModuleName := usedModuleInput.GetStore().ModuleName
+			if packageModulesMapping[inputStoreModuleName].Output.Type != packageModulesMapping[usedModuleStoreMapModuleName].Output.Type {
```
sduchesneau: watch out for panic here, we want to first check if that new module name exists and error out correctly.

manifest/package.go
```diff
@@ -90,6 +98,87 @@ func (r *manifestConverter) validateManifest(manif *Manifest) error {
...
+				return fmt.Errorf("module %q: input %q has different output than the used module %q: input %q", manifestModuleWithUse.Name, input.String(), manifestModuleWithUse.Use, usedModuleInput.String())
+			}
+
+		}
+		if input.GetMap() != nil {
```
sduchesneau: another "else if" here ... all of them!

manifest/package.go
```diff
@@ -90,6 +98,87 @@ func (r *manifestConverter) validateManifest(manif *Manifest) error {
...
+			if usedModuleInput.GetMap() == nil {
+				return fmt.Errorf("module %q: input %q is not a map type", manifestModuleWithUse.Name, input.String())
+			}
+
+			inputMapModuleName := input.GetMap().ModuleName
+			usedModuleInputMapModuleName := usedModuleInput.GetMap().ModuleName
+			if packageModulesMapping[inputMapModuleName].Output.Type != packageModulesMapping[usedModuleInputMapModuleName].Output.Type {
```
sduchesneau: also watch out for panic, first check that the modulename exists in the map

sduchesneau (changes requested):
not bad, few changes requested!

sduchesneau (approved):
LGTM

PR #453, opened 2024-04-05
Feature/index executor

ArnaudBger:
Index Modules and Block Filter

* *Index Modules* and *Block Filter* can now be used to speed up processing and reduce the amount of parsed data.
* When indexes are used along with the `BlockFilter` attribute on a mapper, blocks can be skipped completely: they will not be run in downstreams modules or sent in the output stream, except in live segment or in dev-mode, where an empty 'clock' is still sent.
* See https://github.com/streamingfast/substreams-foundational-modules for an example implementation
* Blocks that are skipped will still appear in the metering as "read bytes" (unless a full segment is skipped), but the index stores themselves are not "metered"


pipeline/exec/indexexec.go
```diff
@@ -0,0 +1,81 @@
...
+func (i *IndexModuleExecutor) Name() string   { return i.moduleName }
+func (i *IndexModuleExecutor) String() string { return i.Name() }
+
+func (i *IndexModuleExecutor) applyCachedOutput([]byte) error {
+	return nil
+}
+
+func (i *IndexModuleExecutor) run(ctx context.Context, reader execout.ExecutionOutputGetter) (out []byte, moduleOutputData *pbssinternal.ModuleOutput, err error) {
+	//TODO: HANDLE exec_index
+	//ctx, span := reqctx.WithModuleExecutionSpan(ctx, "exec_index")
```
sduchesneau: You can add this span, no problem with this commented code

pipeline/exec/indexexec.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	return &pbssinternal.ModuleOutput{
+		Data: &pbssinternal.ModuleOutput_IndexKeys{
+			IndexKeys: &indexKeys,
+		},
+	}, nil
+}
+
+func (i *IndexModuleExecutor) HasValidOutput() bool {
```
sduchesneau, line 67: This should return "false" for now, but the right solution is to get rid of this 'validoutput' mechanism completely.

pipeline/outputmodules/graph.go
```diff
@@ -172,6 +172,13 @@ func computeStages(mods []*pbsubstreams.Module) (stages ExecutionStages) {
 				}
 			}
 
+			//Check block index dependence
```
sduchesneau, line 266: I'll have to double check that with you.. I'm not certain if all cases are covered or if some modules may be pushed down another stage because it wasn't ordered correctly..

pipeline/pipeline.go
```diff
@@ -45,7 +45,7 @@ type Pipeline struct {
 	wasmRuntime     *wasm.Registry
 	outputGraph     *outputmodules.Graph
 	loadedModules   map[uint32]wasm.Module
-	moduleExecutors [][]exec.ModuleExecutor // Staged module executors
+	ModuleExecutors [][]exec.ModuleExecutor // Staged module executors
```
sduchesneau, line 50: Not sure why this became exported ? don't see the usage...

sduchesneau (commented):
just left a few comments, I'm not 100% convinced so not approving until we can talk more!

PR #463, opened 2024-04-25
feat: add `omitempty` in `Manifest` + `MarshalYAML` for `mapSlice`

abhimanyusinghgaur:
We have internal tooling that uses the `Manifest` struct to generate `substreams.yaml` files.
1. Without the `omitempty`, those files have a lot of noise, so adding it.
2. The `mapSlice` type didn't had a `MarshalYAML` defined resulting in it getting marshaled as a slice instead of a map. When reading back a `substreams.yaml` generated this way into the `Manifest` struct, the yaml decoder fails to decode it. So, added the `MarshalYAML` to fix the same.

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -10,8 +10,9 @@ import (
 	"regexp"
 	"strings"
 
-	pbsubstreams "github.com/streamingfast/substreams/pb/sf/substreams/v1"
 	"gopkg.in/yaml.v3"
+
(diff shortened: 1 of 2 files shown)


abhimanyusinghgaur:
Seems like those tests are failing on the HEAD commit in the base branch itself. Please let me know if I need to do anything.

maoueh:
No need to wait, I don't think any test failed because of your change.

PR #470, opened 2024-05-15
storage: fix memory leak when preloading files

colindickson:


storage/execout/filewalk.go
```diff
@@ -71,19 +91,53 @@ func (fw *FileWalker) preload(ctx context.Context, seg int) {
...
+	fw.currentlyPreloadingLock.Unlock()
+
 	f := fw.config.NewFile(rng)
-	go func() {
-		if err := f.Load(ctx); err == nil {
+	fw.buffer[seg] = f
+
+	go func(file *File) {
+		defer func() {
+			fw.currentlyPreloadingLock.Lock()
```
maoueh, line 103: Do we want a read lock here?

storage/execout/filewalk.go
```diff
@@ -71,19 +91,53 @@ func (fw *FileWalker) preload(ctx context.Context, seg int) {
...
+			fw.bufferLock.Lock()
+			fw.currentlyPreloadingLock.Lock()
+
+			delete(fw.buffer, k)
+			delete(fw.currentlyPreloadingSegments, k)
+
+			fw.currentlyPreloadingLock.Unlock()
+			fw.bufferLock.Unlock()
+		}
+	}
```
maoueh, line 140: There is a chance that a key has been added/deleted between those two `fw.bufferLock.Lock()`, what would it cause? 

Should we step lock the full thing here?

maoueh:
This PR seems abandoned, shall I kill it?

colindickson:
yes @maoueh .  it turns out the memory leak was coming from elsewhere.

PR #471, opened 2024-05-16
Add indexes documentation

enoldev:


docs/new/develop/indexes.md
```diff
@@ -0,0 +1,99 @@
+When you execute your Substreams for the first time, you are reading the data stored in the flat files of the Substreams provider.
```
sduchesneau: maybe use "block files" instead of "flat files" to refer to the raw material on first runs.
The "cache" are also "flat" and sometimes we use that term for cached data.

docs/new/develop/indexes.md
```diff
@@ -0,0 +1,99 @@
+When you execute your Substreams for the first time, you are reading the data stored in the flat files of the Substreams provider. 
+
+To improve the performance, the data accessed by the Substreams is cached, so that the second or the third time that you run the Substreams, you can read from the cache, thus saving time and money. This behavior is illustrated in the following diagram.
+
+<figure><img src="../../.gitbook/assets/develop/cache-substreams-flow.png" width="100%" /></figure>
```
sduchesneau, line 5: In that schema for first run, the data is written to the cache AND actually READ FROM CACHE to the user. (the parallel workers create the cache and the "tier1" just reads those cache).
I'm not sure how important it is to specify it on schema, but it is important for me :P

docs/new/develop/indexes.md
```diff
@@ -0,0 +1,99 @@
...
+
+This data indexing is done implicitly every time you run a Substreams for the first time, but Substreams also allows you to explicitly specify which specific data you want to index for further performance improvement.
```
sduchesneau: > This data indexing
I would rather say: " This data _caching_ " (not indexing! even if this term is sometimes used that way... it will create confusion here)

> allows you to explicitly specify which specific data you want to index
I would rather say:
"allows you to create an additional index on top of your data"
or something like this, to emphasize that the indexes are "different" than simply caching the mapper data.

docs/new/develop/indexes.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Indexes
+
+Substreams has recently introduced the concept of _indexed modules_. An indexed module is a module that has been pre-cached for some specific data. Let's see it with an example!
```
sduchesneau: I think "index modules" is clearer than "indexed modules".
We are not creating a module that happens to be indexed... We are creating an index! the module is the definition of the index!

docs/new/develop/indexes.md
```diff
@@ -0,0 +1,99 @@
...
+
+With indexing, you could have a pre-cached module with the information of all the event addresses in the block. Instead of reading the full Ethereum block, you can search in the events index (event cache) and avoid decoding the data of those blocks that do not contain events you are interested in.
+
+In the following diagram, you can see three blocks with their corresponding data. Consider that you want to retrieve all the events where `log.address == 0xcd2...`. In a non-indexed module, you would have to decode the data of every block, iterate over the logs, and search if there is any log matching the condition.
+
+<figure><img src="../../.gitbook/assets/develop/cache-block-1.png" width="100%" /></figure>
+
+On the other hand, in an indexed module, the events of every block are pre-cached in a special store, so when you look for events where `log.address == 0xcd2...`, you can simply search in the index store of the block. If the event is contained, within the block, then you decode the data. If not, you skip it.
```
sduchesneau: in the following few paragraphs, there is confusion on two specific concepts:
1) the precaching of the "events" module
2) the index module that indicates where each event is contained...
I don't think it's ok as it is.


"In a non-indexed module, you would have to..." should be -> "without an index, you would have to go through the data of every block, but with an index, you can skip the blocks that do not contain the event that you want"

docs/new/develop/indexes.md
```diff
@@ -0,0 +1,99 @@
...
+
+In the following diagram, `Block 1` and `Block 2` contain an event where `log.address == 0xcd2...`, but `Block 3` does not.
+
+<figure><img src="../../.gitbook/assets/develop/cache-block-2.png" width="100%" /></figure>
+
+### Create a Custom Index
+
+Anyone can create an indexed module. All you need to do is create a Substreams that indexes the piece of data of your choice. For example, let's take a look at the `index_events` module from the [Ethereum Foundational Modules GitHub repository](https://github.com/streamingfast/substreams-foundational-modules/blob/develop/ethereum-common/substreams.yaml#L53).
```
sduchesneau: >create an indexed module
-> create an "index" module.

> create a Substreams that indexes the piece of data of your choice. 
-> create a Substreams with a module that outputs the list of terms that are contained in each block.

docs/new/develop/indexes.md
```diff
@@ -0,0 +1,99 @@
...
+
+The flow of indexing the events of a block looks like the following:
```
sduchesneau: The following flow is a good, as it explains a "good way to use an index" (in conjunction with: a first module that reduces the amount of data and a 3rd module that filters the events found in the blocks matching the index)

We should insist that this is one usage pattern for index modules, it is not the only way of using an index!
again, this results from the confusion between the role of each of the 3 modules in there...

docs/new/develop/indexes.md
```diff
@@ -0,0 +1,99 @@
...
+
+<figure><img src="../../.gitbook/assets/develop/cache-flow.png" width="100%" /></figure>
+
+1. The `all_events` module receives a `Block` object as an input and outputs an `Events`, with all the events of the block.
+2. The `index_events` module receives the `Events` object as an input and outputs a `Keys` object, containing the `address` and `signature` fields of every event you want to track.
+3. The `filtered_events` receives the `index_events` module as an input plus a string with the event addresses that the Substreams must filter.
+Given this string of addresses, Substreams checks if the event address is contained on a given block before actually decoding the data of the block.
+You can use logical operators (`and` and `or`) to select what events to search.
+
+The definition of the `index_events` module looks like any other Substreams module, but it is a special _kind_, `kind: blockIndex` and outputs a special data model, `sf.substreams.index.v1.Keys`. The `Keys` object contains the data that must be cached.
```
sduchesneau: > The `Keys` object contains the data that must be cached.
would be clearer to explain as:
-> contains a list of labels that will be used to identify the content of that block

docs/new/develop/indexes.md
```diff
@@ -0,0 +1,99 @@
...
+
+The keys of the block, defined by the `Keys` object, are a list of strings defining the parts of the event that you want to use for searching. For example:
+
+```
+Block 32443
+--------------------------
+keys = {'evt_addr:0xa34', 'evt_addr:0xba7', 'evt_addr:0x99a'}
+```
+
+If you're looking for an event with address `0xba7`, when Substreams gets to block this block, it will know beforehand that the block contains that event. If you looking for an event with address `0xaa1`, then Substreams knows beforehand it's not contained in the block and can safely skip it.
```
sduchesneau: > block this block
typo !
the rest is good...

sduchesneau (changes requested):
clarify confusion between 
* the "index module", 
* the mapper module that reduces full blocks to list of events (reducing module?)
* the mapper module that applies the filter query to the list of events to further reduce the content (filtering module?)

PR #476, opened 2024-05-29
This enables running module with wasmbindgen shims by specifying runtime extension `wasm-bindgen-shims`

maoueh:
The Substreams binary type can now be optionally composed of runtime extensions by appending a `+<extension>,[<extesions...>]` at the end of the binary type. Extensions are `key[=value]` that are runtime specifics. If you were parsing generic Substreams manifest, you will now need to handle that possibility in the binary type. If you were reading the field without any processing, you don't have to change nothing.

The first runtime extensions that we add is `wasm-bindgen-shims` which will enable Substreams importing those shims to run correctly as long as they are never called which is usually the case. We achieve this my binding wasm bindgen module and their imports to a dummy implementation that fails as soon as it's called.

This enables Substreams to import for example `alloy` in the Ethereum world to perform some Ethereum related operations. It also enables developer to pull in `solana_program` in Solana world so they are able to decode IDL directly.

To enable the feature, change:

```
binaries:
  default:
    type: wasm/rust-v1
    file: <some_file>
```

to become

```
binaries:
  default:
    type: wasm/rust-v1+wasm-bindgen-shims
    file: <some_file>
```

This instructs the Substreams processing engine on the server to shims the WASM bindgen imports with dummy implementation.


wasm/wazero/module.go
```diff
@@ -231,3 +284,17 @@ func AddHostFunctions(ctx context.Context, runtime wazero.Runtime, moduleName st
...
 	return build.Compile(ctx)
 }
+
+func addUnboundedModule(ctx context.Context, runtime wazero.Runtime, moduleName string, imports map[string]api.FunctionDefinition) (wazero.CompiledModule, error) {
+	build := runtime.NewHostModuleBuilder(moduleName)
+
+	for importName, f := range imports {
+		build.NewFunctionBuilder().
+			WithGoFunction(api.GoFunc(func(ctx context.Context, stack []uint64) {
+				panic(fmt.Errorf("you are trying to call %q from module %q for which Substreams provided a dummy"+
```
abourget, line 294: is there not a better way to add logs to the module's output?
Does that bubble up through the gRPC connection properly?

wasm/extensions.go
```diff
@@ -0,0 +1,56 @@
...
+			valuePtr = &value
+		}
+
+		extensions = append(extensions, RuntimeExtension{
+			ID:    RuntimeExtensionID(id),
+			Value: valuePtr,
+		})
+	}
+
+	return
```
abourget, line 55: hehe sophisticated :)

abourget: was that necessary?

wasm/wazero/module.go
maoueh: This does trickles down to the gRPC connection

wasm/extensions.go
maoueh: Better defined a strict output than be lenient. Doesn't hurt and typos are properly caught.

PR #492, opened 2024-06-10
Feature/live back filling

ArnaudBger:


sduchesneau:
LGTM, just missing the CHANGELOG entry and add some comments on the tests to explain why it works with these numbers: it isn't obvious from an outsider point of view

PR #518, opened 2024-08-16
Improvements to the documentation

enoldev:


docs/new/references/chains-and-endpoints.md
```diff
@@ -31,9 +31,24 @@ Protobuf definitions and public endpoints are provided for the supported protoco
...
 | Antelope  | [`sf.antelope.type.v1.Block`](https://buf.build/pinax/firehose-antelope/docs/main:sf.antelope.type.v1#sf.antelope.type.v1.Block)              |                                                                                                                       |
+| Starknet  | [`sf.starknet.type.v1.Block`](https://github.com/streamingfast/firehose-starknet/blob/main/proto/sf/starknet/type/v1/block.proto#L8)              |                                                                                                                       |
+| Gara  | [`sf.gear.type.v1.Block`](https://github.com/streamingfast/firehose-gear/blob/main/proto/sf/gear/type/v1/block.proto#L7)              |                                                                                                                       |
+
+### EVM Extended vs Base Block Model
+
+For EVM chains, the are two different types of `Block` models: _Extended_ and _Base_:
```
maoueh, line 42: One comment I would add here is that we should note that both model still uses the same Protobuf message, just that in extended the `Block` is more populated than the in `base` but they both use the same structure.

PR #523, opened 2024-08-26
Add Arbitrum & Base to manifest chains

enoldev:


sduchesneau (changes requested):
Watch out for alignment.

You can (should!!) use auto-formatting when coding in go (either as an option in your editor, or running the `gofmt` command on your file)

Set this up and fix the alignement! then all good, of course

sduchesneau (approved):

PR #574, opened 2025-01-28
Limit parallel execution of a stage's layer

maoueh:
Previously, the engine was executing modules in a stage's layer all in parallel. So if you had 20 independent mapper modules, they were all run in parallel.

This was hindering performance on high load where a lot of CPU cycles can be consumed will the machine has limited physical cores available.

We now change that behavior, development mode will not execute any modules in parallel, never. For production mode, we now limit to 2 parallel execution. A future update will make that value dynamic based on the subscription of the request.

### DIFF
--- a/pipeline/process_block.go
+++ b/pipeline/process_block.go
@@ -6,11 +6,11 @@ import (
 	"fmt"
 	"io"
 	"runtime/debug"
-	"sync"
 
 	"github.com/streamingfast/bstream"
 	pbbstream "github.com/streamingfast/bstream/pb/sf/bstream/v1"
 	"github.com/streamingfast/dmetering"
+	"github.com/streamingfast/logging"
 	"github.com/streamingfast/substreams/metering"
 	"github.com/streamingfast/substreams/metrics"
 	pbssinternal "github.com/streamingfast/substreams/pb/sf/substreams/intern/v2"
@@ -20,6 +20,7 @@ import (
 	"github.com/streamingfast/substreams/reqctx"
 	"github.com/streamingfast/substreams/storage/execout"
 	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/protobuf/types/known/anypb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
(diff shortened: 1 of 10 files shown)


service/tier1.go
```diff
@@ -433,18 +433,22 @@ func (s *Tier1Service) blocks(ctx context.Context, request *pbsubstreamsrpc.Requ
 	cacheTag := s.runtimeConfig.DefaultCacheTag
 	if auth := dauth.FromContext(ctx); auth != nil {
 		if parallelJobs := auth.Get("X-Sf-Substreams-Parallel-Jobs"); parallelJobs != "" {
-			if ll, err := strconv.ParseUint(parallelJobs, 10, 64); err == nil {
-				requestDetails.MaxParallelJobs = ll
+			if count, err := strconv.ParseUint(parallelJobs, 10, 64); err == nil {
+				requestDetails.MaxParallelJobs = count
+			}
+		}
+		if parallelExecutors := auth.Get("X-Sf-Substreams-Stage-Layer-Parallel-Executor-Max-Count"); parallelExecutors != "" {
```
sduchesneau: need to:
1) implement the same parsing in tier2.go
2) ensure we pass that header down to tier2 request

sduchesneau (changes requested):
just ensure tier2 gets the same treatment passed down from tier1, and the rest is very good!

service/tier1.go
maoueh: Done please check

reqctx/context.go
```diff
@@ -91,6 +91,40 @@ func WithEmitter(ctx context.Context, emitter dmetering.EventEmitter) context.Co
...
+	}
+
+	return !details.ProductionMode
+}
+
+const defaultMaxStageLayerParallelExecutorCount = 2
+
+const safeguardMaxStageLayerParallelExecutorCount = 16
+
+// MaxParallelJobs returns the maximum number of parallel executors (e.g. go routines) that can
```
sduchesneau: typo, should be MaxStageLayerParallelExecutor

pipeline/process_block.go
```diff
@@ -307,11 +308,14 @@ func (p *Pipeline) executeModules(ctx context.Context, execOutput execout.Execut
...
-		if len(stage) < 2 {
-			//fmt.Println("Linear stage", len(stage))
-			for _, executor := range stage {
+
+	isDevelopmentMode := reqctx.IsInDevelopmentMode(ctx)
+	maxParallelExecutor := reqctx.MaxStageLayerParallelExecutor(ctx)
+	logging.Logger(ctx, p.stores.logger).Debug("executing stage's layers", zap.Int("layer_count", len(p.StagedModuleExecutors)), zap.Uint64("max_parallel_executor", maxParallelExecutor))
+
+	for _, layer := range p.StagedModuleExecutors {
+		if isDevelopmentMode || maxParallelExecutor <= 1 || len(layer) <= 1 {
```
sduchesneau: why do we care here about isDevelopmentMode ?

The `MaxStageLayerParallelExecutor` should check that we are in development mode and return 1. This logic doesn't belong here.

reqctx/context.go
```diff
@@ -91,6 +91,40 @@ func WithEmitter(ctx context.Context, emitter dmetering.EventEmitter) context.Co
...
+// If the value is not set, it will return 1.
+func MaxStageLayerParallelExecutor(ctx context.Context) uint64 {
+	details := Details(ctx)
+	if details == nil {
+		// Always give at least 1, but there should always be a details object attached to the context
+		return 1
+	}
+
+	// If unset, provide default value which is 2 for now
+	if details.MaxStageLayerParallelExecutor == 0 {
```
sduchesneau, line 117: as mentionned earlier, add the check for IsInDevelopmentMode here and return 1 ...

sduchesneau (changes requested):
typo in comment, and I suggest moving the IsInDevelopment logic to the code evaluating the max number of workers for proper separation of concerns.

Other than that, LGTM !

PR #576, opened 2025-02-03
Add protocol selection in the "substreams init" command

enoldev:


proto/sf/codegen/conversation/conversation.proto
```diff
@@ -0,0 +1,162 @@
+syntax = "proto3";
```
maoueh: Do we need that file actually?

cmd/substreams/init.go
```diff
@@ -172,8 +172,63 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
+			case "cosmos":
+				return "Cosmos"
+			case "starknet":
+				return "Starknet"
+			}
+
+			return ""
+		}
+
+		var filteredProtocols map[string]*BlockchainProtocolSelector = make(map[string]*BlockchainProtocolSelector)
```
maoueh:
```suggestion
	filteredProtocols := make(map[string]*BlockchainProtocolSelector)
```

cmd/substreams/init.go
```diff
@@ -172,8 +172,63 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
+			} else {
+				filteredProtocols[gen.Group] = &BlockchainProtocolSelector{
+					Id:         gen.Group,
+					Title:      titleFromGroup(gen.Group),
+					Generators: []*pbconvo.DiscoveryResponse_Generator{gen},
+				}
+			}
+		}
+
+		var protocolOptions []huh.Option[*BlockchainProtocolSelector]
```
maoueh: It's good practice to pre-allocate the array when you know it's fixed size, like here.
```suggestion
	protocolOptions := make([]huh.Option[*BlockchainProtocolSelector], 0, len(filteredProtocols))
```

cmd/substreams/init.go
```diff
@@ -172,8 +172,63 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
+			Value(&selector)
+
+		err = huh.NewForm(huh.NewGroup(selectField)).WithTheme(huhTheme).WithAccessible(WITH_ACCESSIBLE).Run()
+		if err != nil {
+			return fmt.Errorf("failed taking input: %w", err)
+		}
+
+		fmt.Println(gray("┃"), "Chosen chain: ", bold(selector.Id), "-", selector.Title)
+
+		var generatorOptions []huh.Option[*pbconvo.DiscoveryResponse_Generator]
```
maoueh:
```suggestion
	generatorOptions := make([]huh.Option[*pbconvo.DiscoveryResponse_Generator], 0, len(selector.Generators))
```

cmd/substreams/init.go
```diff
@@ -188,21 +243,23 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
 				Key:   key,
 				Value: gen,
 			}
-			options = append(options, entry)
+			generatorOptions = append(generatorOptions, entry)
 		}
 
+		fmt.Printf("Testing................")
```
maoueh: Leftover?

cmd/substreams/init.go
```diff
@@ -188,21 +243,23 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
 		var codegen *pbconvo.DiscoveryResponse_Generator
-		selectField := huh.NewSelect[*pbconvo.DiscoveryResponse_Generator]().
-			Title("Choose the code generator that you want to use to bootstrap your project").
-			Options(options...).
+		selectGeneratorField := huh.NewSelect[*pbconvo.DiscoveryResponse_Generator]().
+			Title("Choose the chain ecosystem that you want to use to bootstrap your project (EVM, Cosmos, Solana, etc)").
```
maoueh: What the difference with `Choose the protocol you inted to work with:`? Isn't this question to choose a specific generator, even now?

cmd/substreams/init.go
```diff
@@ -172,8 +172,63 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
+			Title("Choose the protocol you inted to work with:").
+			Options(protocolOptions...).
+			Value(&selector)
+
+		err = huh.NewForm(huh.NewGroup(selectField)).WithTheme(huhTheme).WithAccessible(WITH_ACCESSIBLE).Run()
+		if err != nil {
+			return fmt.Errorf("failed taking input: %w", err)
+		}
+
+		fmt.Println(gray("┃"), "Chosen chain: ", bold(selector.Id), "-", selector.Title)
```
maoueh: You've used `protocol` in the question but `Chosen chain` here, shouldn't it be `Chosen protocol: `?

cmd/substreams/init.go
```diff
@@ -172,8 +172,63 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("failed to call discovery endpoint: %w", err)
 		}
 
-		var options []huh.Option[*pbconvo.DiscoveryResponse_Generator]
+		type BlockchainProtocolSelector struct {
+			Id         string
+			Title      string
+			Generators []*pbconvo.DiscoveryResponse_Generator
+		}
+		titleFromGroup := func(group string) string {
```
maoueh: I think this should be extracted to it's own function, to reduce the clutter in the `runSubstreamsInit` a bit. If extracted, I would name it `generatorGroupToProtocolTitle(...)`

maoueh (changes requested):

maoueh (approved):

maoueh:
There is just a small conflict, rebase and re-generate the Protobuf, and we should be good

Omarturb (approved):

PR #579, opened 2025-02-05
shared WASM execution cache

sduchesneau:
adds a cache for module*block executed near the HEAD of the chain, to prevent executing the same foundational modules 5 times whenever a block comes in


app/tier1.go
```diff
@@ -67,6 +77,7 @@ type Tier1Config struct {
 	SubrequestsEndpoint  string
 	SubrequestsInsecure  bool
 	SubrequestsPlaintext bool
+	SharedCacheSize      uint64
```
maoueh, line 80: Size in what? Bytes, count, etc?

metrics/metrics.go
```diff
@@ -37,6 +37,9 @@ var Tier2RejectedRequestCounter = MetricSet.NewCounterVec(
 	"Counter for total Substreams requests the tier2 rejected, by reason (gRPC code reason)",
 )
 
+var ExecutedWasmModules = MetricSet.NewCounter("substreams_executed_wasm_modules", "Counter for total Substreams modules x block for which wasm was executed")
```
maoueh: I don't get the `... x block for` part

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+
+type SharedCache struct {
+	sync.Mutex
+	headBlock *atomic.Uint64
+	size      uint64
+	// module_hash -> blockHash -> callEntry
+	callEntries map[string]map[string]*callEntry
+}
+
+const maxUint64 uint64 = 1<<64 - 1
```
maoueh: Not useful, `math.MaxUint64` works out of the box

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+
+func NewSharedCache(size uint64) *SharedCache {
+	return &SharedCache{
+		headBlock:   atomic.NewUint64(maxUint64),
```
maoueh:
```suggestion
		headBlock:   atomic.NewUint64(math.MaxUint64),
```

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+	"go.uber.org/zap"
+)
+
+var GlobalSharedCache *SharedCache
+
+type SharedCache struct {
+	sync.Mutex
+	headBlock *atomic.Uint64
+	size      uint64
+	// module_hash -> blockHash -> callEntry
```
maoueh, line 23: Need the "kind" of size, I think it's in `block`.

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+	res.executionStack = append([]string{}, call.ExecutionStack...)
+	res.executed = true
+}
+
+func (s *SharedCache) Cachable(blockNum uint64) bool {
+	resp := s != nil && blockNum+s.size > s.headBlock.Load()
+	return resp
+}
+
+func (s *SharedCache) Execute(
```
maoueh, line 113: There is a Golang warning here:

```
argument ctx is overwritten before first use (SA4009)
	sharedcache.go:163:2: assignment to ctxgo-staticcheck
```

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+	for clockID, hashesInClock := range s.callEntries {
+		for entryID, entry := range hashesInClock {
+			if entry.clock.Number < lowest {
+				delete(hashesInClock, entryID)
+				if tracer.Enabled() {
+					zlog.Debug("deleting entry", zap.Uint64("clock_num", entry.clock.Number))
+				}
+			}
+		}
+		if len(hashesInClock) == 0 {
```
maoueh: When is it possible that this becomes 0 actually?

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+
+	err      error
+	panicErr *wasm.PanicError
+}
+
+func applyResult(res *callEntry, call *wasm.Call) error {
+	if call.Clock.Id != res.clock.Id ||
+		call.Entrypoint != res.entrypoint ||
+		call.ModuleName != res.moduleName {
+		panic(fmt.Sprintf("invalid shared cache data on block %s (%s) for module %s (%s)", call.Clock, res.clock, call.ModuleName, res.moduleName))
```
maoueh: I understand this should not happen, but since we already have a returned error here, wouldn't it better to not crash everything, what do you think?

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+	res.returnValue = call.Output()
+	res.logs = append([]string{}, call.Logs...)
+	res.logsByteCount = call.LogsByteCount
+	res.executionStack = append([]string{}, call.ExecutionStack...)
+	res.executed = true
+}
+
+func (s *SharedCache) Cachable(blockNum uint64) bool {
+	resp := s != nil && blockNum+s.size > s.headBlock.Load()
+	return resp
```
maoueh:
```suggestion
	return s != nil && blockNum+s.size > s.headBlock.Load()
```

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+	clock := call.Clock
+	var resultLockOwner bool
+
+	s.Lock()
+
+	if s.callEntries[clock.Id] == nil {
+		s.callEntries[clock.Id] = make(map[string]*callEntry)
+	}
+	result, ok := s.callEntries[clock.Id][moduleHash]
+	if !ok {
```
maoueh: `found` is almost always clearer than `ok` for a map lookup
```suggestion
	result, found := s.callEntries[clock.Id][moduleHash]
	if !found {
```

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+			zlog.Debug("getting wasm call from cache", zap.String("module_hash", moduleHash), zap.Uint64("block_num", clock.Number))
+		}
+		metrics.SkippedCachedWasmModules.Inc()
+		return applyResult(result, call)
+	}
+
+	if tracer.Enabled() {
+		zlog.Debug("executing wasm call", zap.String("module_hash", moduleHash), zap.Uint64("block_num", clock.Number))
+	}
+	metrics.ExecutedWasmModules.Inc()
```
maoueh: Best would be to link that to the app context, so at least it would to be ever blocking

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+	s.Lock()
+
+	if s.callEntries[clock.Id] == nil {
+		s.callEntries[clock.Id] = make(map[string]*callEntry)
+	}
+	result, ok := s.callEntries[clock.Id][moduleHash]
+	if !ok {
+		result = &callEntry{clock: clock}
+		s.callEntries[clock.Id][moduleHash] = result
+		resultLockOwner = true
```
maoueh: It would be good to add a comment here to explain the overall process that this is going to block other modules that arrives after this one and that it's a write lock because it's going to actually write the results.

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+		result.Lock()
+		defer result.Unlock()
+	}
+
+	s.Unlock()
+
```
maoueh: Maybe it would be better to put that in the `else` clause above, tied it more closely where it matters.

pipeline/exec/sharedcache.go
```diff
@@ -0,0 +1,170 @@
...
+func (s *SharedCache) Execute(
+	ctx context.Context,
+	wasmModule wasm.Module,
+	moduleHash string,
+	call *wasm.Call,
+	wasmArguments []wasm.Argument,
+	argValues map[string][]byte,
+) error {
+	clock := call.Clock
+	var resultLockOwner bool
```
maoueh:
```suggestion
// Manually unlocked, pay attention if you add early exit beyond that lock call
```

service/tier1.go
```diff
@@ -189,6 +190,22 @@ func NewTier1(
...
+		}
+		<-hub.Ready
+		sharedCache := exec.NewSharedCache(sharedCacheSize)
+		hubSrc := hub.SourceFromBlockNum(hub.HeadNum(), sharedCache)
+		if hubSrc == nil {
+			zlog.Error("cannot get blocks source from hub")
+			return
+		}
+		exec.GlobalSharedCache = sharedCache
+		hubSrc.Run()
```
maoueh, line 206: Should we check for errors here, at least for logging purposes?

maoueh (changes requested):

pipeline/exec/sharedcache.go
maoueh: We could leverage this `ctx` to extract a logger, so you would get trace ids correctly. I understand we do not use it to complete execution for others.

PR #583, opened 2025-02-14
Update default GUI bytes representation when using Solana

enoldev:


tui2/pages/output/output.go
```diff
@@ -72,6 +72,16 @@ func New(c common.Common, config *request.Config) (*Output, error) {
 	// 	return nil, err
 	// }
 
+	getBytesEncodingPerNetwork := func(endpoint string) dynamic.BytesRepresentation {
+		if manifest.HardcodedEndpoints["solana-mainnet-beta"] == endpoint {
+			return dynamic.BytesAsBase58
+		}
+
+		return dynamic.BytesAsBase64
```
maoueh: Can you default to hex instead, seems like a better default for most chains for now. 

The network registry from Pinax was planned to be integrated which should come with bytes encoding.

maoueh (approved):

tui2/pages/output/output.go
```diff
@@ -72,6 +72,16 @@ func New(c common.Common, config *request.Config) (*Output, error) {
 	// 	return nil, err
 	// }
 
+	getBytesEncodingPerNetwork := func(endpoint string) dynamic.BytesRepresentation {
+		if manifest.HardcodedEndpoints["solana-mainnet-beta"] == endpoint {
```
maoueh: I thought about this, do you want to also specify `solana-devnet` here?

codegen/chainconfig.go
```diff
@@ -134,4 +134,10 @@ var ChainConfigByID = map[string]*ChainConfig{
 		FirehoseEndpoint: "mainnet.stellar.streamingfast.io:443",
 		Network:          "stellar-mainnet",
 	},
+	"sei-mainnet": {
+		DisplayName:      "Sei Mainnet",
+		ExplorerLink:     "",
```
maoueh: Don't we want to put seitrace here maybe?

codegen/chainconfig.go
```diff
@@ -134,4 +134,10 @@ var ChainConfigByID = map[string]*ChainConfig{
 		FirehoseEndpoint: "mainnet.stellar.streamingfast.io:443",
 		Network:          "stellar-mainnet",
 	},
+	"sei-mainnet": {
+		DisplayName:      "Sei Mainnet",
```
maoueh, line 138: How to they write it, `SEI Mainnet` or `Sei Mainnet`?

enoldev: In the documentation, they always seem to use `Sei`
https://www.docs.sei.io/learn/general-overview

maoueh (approved):

PR #588, opened 2025-02-27
Add Local File reading in substreams init

enoldev:


cmd/substreams/init.go
```diff
@@ -419,6 +419,39 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
+			content, err := os.ReadFile(returnValue)
+			if err != nil {
+				return errors.New("could not read the file provided")
+			}
+
+			if err := sendFunc(&pbconvo.UserInput{
+				FromActionId: resp.ActionId,
+				Entry: &pbconvo.UserInput_LocalFile_{
+					LocalFile: &pbconvo.UserInput_LocalFile{Value: strings.TrimRight(string(content), " ")},
+				},
```
maoueh: We never want to treat local file as text, it needs to stay in `[]byte` form.

proto/sf/codegen/conversation/v1/conversation.proto
```diff
@@ -24,10 +24,14 @@ message UserInput {
     Upload file = 16;
     // Deprecated: we don't use this.
     DownloadedFiles downloaded_files = 20;
+    LocalFile local_file = 21;
   }
   message TextInput {
     string value = 1;
   }
+  message LocalFile {
+    string value = 1;
```
maoueh: `bytes`

maoueh (changes requested):

maoueh (approved):

PR #604, opened 2025-03-21
Add Stellar to the init command

enoldev:


cmd/substreams/init.go
```diff
@@ -515,23 +530,61 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
+					savingDest = projectName
+				}
+
+				if cwd, err := os.Getwd(); err == nil {
+					savingDest = filepath.Join(cwd, savingDest)
+				}
+
+				inputField := huh.NewInput().Title("In which directory do you want to download the project?").Value(&savingDest)
+				inputField.Validate(func(userInput string) error {
+					fmt.Println("Checking directory", userInput)
```
maoueh, line 545: Leftover?

maoueh: I think it's part of the UI

cmd/substreams/init.go
```diff
@@ -515,23 +530,61 @@ func runSubstreamsInitE(cmd *cobra.Command, args []string) error {
...
-			savingDest, _ := os.Getwd()
+			if userState.downloadedFilesfolderPath == "" {
+				savingDest := "output"
+				if projectName := gjson.GetBytes(lastState.State, "name").String(); projectName != "" {
+					savingDest = projectName
+				}
+
+				if cwd, err := os.Getwd(); err == nil {
+					savingDest = filepath.Join(cwd, savingDest)
+				}
```
maoueh, line 541: Would have been better to use `filepath.Abs` which does this already and much more.

PR #616, opened 2025-04-27
Add network & networks properties

vinhloc30796:


schemas/manifest-schema.json
```diff
@@ -23,8 +23,433 @@
         "version": {
           "title": "package version",
           "description": "A package version\nhttps://substreams.streamingfast.io/reference-and-specs/manifests#package.version",
+          "type": "string"{
```
maoueh: Is that some typo?

maoueh (changes requested):
Unsure if this was done right, the JSON seems malformed now.

schemas/manifest-schema.json
vinhloc30796: fixed now, thanks for catching!

maoueh (approved):

schemas/manifest-schema.json
```diff
@@ -374,6 +374,50 @@
       },
       "additionalItems": false,
       "minItems": 1
+    },
+    "network": {
+      "title": "network",
+      "description": "Target blockchain where the Substreams package will execute. See supported chains at https://thegraph.com/docs/en/supported-networks/",
+      "type": "string",
+      "pattern": "arbitrum-one|avalanche|base|bsc|celo|eos|mainnet|gnosis|near-mainnet|optimism|matic|polygon-zkevm|solana-mainnet-beta"
```
maoueh, line 382: There is probably an ever increasing set of network here, unsure if should restricted to a given set. Would need to be generated from The Graph Network Registry at least.

Path with least resistance is to leave it free form for now.

Wdyt?

schemas/manifest-schema.json
```diff
@@ -374,6 +374,50 @@
...
+    },
+    "networks": {
+      "title": "networks",
+      "description": "Per-network overrides for initialBlock and params",
+      "type": "object",
+      "propertyNames": {
+        "pattern": "arbitrum-one|avalanche|base|bsc|celo|eos|mainnet|gnosis|near-mainnet|optimism|matic|polygon-zkevm|solana-mainnet-beta"
```
maoueh, line 389: Same here.

maoueh (changes requested):

maoueh:
Merging, I'll change the `pattern` to free form though for now.

PR #624, opened 2025-05-08
v8 integration with benchmarks

UlysseCorbeil:


service/validate.go
```diff
@@ -93,6 +93,7 @@ func validateBinaryTypes(bins []*pbsubstreams.Binary) error {
 
 		switch wasmCodeTypeID {
 		case "wasm/rust-v1":
+		case "javascript-v8":
```
maoueh: Let's use something more inline with current formatting:

```suggestion
		case "javascript/v8":
```

wasm/javascript-v8/benchmark.md
```diff
@@ -0,0 +1,39 @@
+# Substreams Javascript runtime integration
```
maoueh: (Comment on package name)

Let's refactor so package location is `wasm/v8` instead of `wasm/javascript-v8`

wasm/javascript-v8/factory.go
```diff
@@ -0,0 +1,19 @@
...
+	"context"
+
+	"github.com/streamingfast/substreams/wasm"
+	"rogchap.com/v8go"
+)
+
+type V8ModuleFactory struct{}
+
+func NewModule(ctx context.Context, wasmCode []byte, wasmCodeType string, registry *wasm.Registry) (wasm.Module, error) {
+	iso := v8go.NewIsolate()
```
maoueh: What does `Isolate` do? A comment would be welcome to describe quickly the implications.

wasm/javascript-v8/instance.go
```diff
@@ -0,0 +1,28 @@
+package v8
+
+import (
+	"context"
+
+	"rogchap.com/v8go"
+)
+
+type V8Instance struct {
```
maoueh: No need to prefix all your struct and method with `V8`, since the package is itself named `v8`, when it will be used it would look like `v8.NewV8Instance` while `v8.NewInstance()` is as descriptive in the context where it's read.

Remove `V8` inside public method within the `v8` package.

wasm/javascript-v8/instance.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func NewV8Instance(iso *v8go.Isolate) (*V8Instance, error) {
+	ctx := v8go.NewContext(iso)
+	return &V8Instance{iso, ctx}, nil
+}
+
+func (inst *V8Instance) Cleanup(ctx context.Context) error {
+	return nil
+}
```
maoueh: What is cleanup doing? I mean, I understand it's coming from the interface, but doing nothing requires a comment here to explain why there is no cleanup to do.

wasm/javascript-v8/instance.go
```diff
@@ -0,0 +1,28 @@
...
+
+func (inst *V8Instance) Close(ctx context.Context) error {
+	inst.ctx.Close()
+	inst.ctx = nil
+	inst.iso = nil
```
maoueh: Not really necessary to reset the structure, usually it will be garbage collected anyway soon after.

wasm/javascript-v8/integration.go
```diff
@@ -0,0 +1,9 @@
+package v8
+
+import (
+	"github.com/streamingfast/substreams/wasm"
+)
+
+func init() {
+	wasm.RegisterModuleFactory("javascript-v8", wasm.ModuleFactoryFunc(NewModule))
```
maoueh:
```suggestion
	wasm.RegisterModuleFactory("javascript/v8", wasm.ModuleFactoryFunc(NewModule))
```

wasm/javascript-v8/module.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if _, err = inst.ctx.RunScript(polyfillCode, "polyfill.js"); err != nil {
+		inst.Close(ctx)
+		return nil, fmt.Errorf("executing polyfill: %w", err)
+	}
+
+	if _, err = inst.ctx.RunScript(preludeCode, "prelude.js"); err != nil {
+		inst.Close(ctx)
+		return nil, fmt.Errorf("executing prelude: %w", err)
+	}
```
maoueh: Would it make sense to concatenate the bundle thus making a single call to `RunScript`?

wasm/javascript-v8/module.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if _, err = inst.ctx.RunScript(string(mod.code), "bundle.js"); err != nil {
+		inst.Close(ctx)
+		return nil, fmt.Errorf("executing JS bundle: %w", err)
+	}
+
+	return inst, nil
```
maoueh: Who is closing the `inst` when there was no error within the script?

Golang as a `defer` functionality to perform an operation after the function finish but before returning to the caller. You duplicated the `inst.Close(...)` call which is error prone. 

Instead, remove all the manual call, and at line 44, do:

```go
		inst, err = NewV8Instance(mod.iso)
		if err != nil {
			return nil, fmt.Errorf("creating new V8 instance: %w", err)
		}
		defer func () { if err := inst.Close(); err != nil { /* log error */} }()
```

wasm/javascript-v8/runtime/prelude.bundle.js
```diff
@@ -0,0 +1,3724 @@
...
+    const out = map_block(readInput());
+    writeOutput(out);
+  }
+  function readInput() {
+    return globalThis.input;
+  }
+  function writeOutput(output) {
+    globalThis.output.set(output);
+  }
+  function map_block(data) {
```
maoueh: Why the prelude contains such code?

wasm/registry.go
```diff
@@ -83,6 +83,7 @@ func NewRegistry(extensions map[string]map[string]WASMExtension) *Registry {
 		"default":                         runtimes[defaultRuntime],
 		"wasip1/tinygo-v1":                runtimes["wazero"], // only wazero supports tinygo at the moment
 		"wasm/rust-v1":                    runtimes[defaultRuntime],
+		"javascript-v8":                   runtimes["javascript-v8"],
```
maoueh:
```suggestion
		"javascript/v8":                   runtimes["javascript/v8"],
```

maoueh (changes requested):

wasm/javascript-v8/instance.go
UlysseCorbeil: Well NewInstance is already used in module.go as per the nomenclature from other wasm modules but its simply calling NewV8Module so it's not confusing. ie: NewInstance calls NewInstance. But I could move the NewInstance logic into module.go and just leave instance.go to do the cleaning up.

wasm/javascript-v8/instance.go
UlysseCorbeil: True but my intentions were simply to make it explicit that the object was unusable at that point and ensure its impossible to call a closed instance. Should I remove it or just add a comment ?

wasm/javascript-v8/module.go
UlysseCorbeil: It does and it will, right now I left them separated for testing purposes since it was easier to see what crashes and where exactly it was coming from

wasm/javascript-v8/runtime/prelude.bundle.js
UlysseCorbeil: For testing purposes, it will not be in the "final" build

wasm/javascript-v8/instance.go
maoueh: This tends to rot, people forget to update `Close` when new fields are added. A comment in the `Close` method that describe that the instance should not be re-used should be enough.

maoueh: So remove

wasm/javascript-v8/module.go
maoueh: After review, the instance is used and lifecycle is handled later, this comment does not appy.

wasm/javascript-v8/runtime/prelude.bundle.js
maoueh: Ensure that the benchmark are run against minimum required prelude+polyfill for now.

cmd/substreams/build.go
```diff
@@ -199,6 +199,9 @@ func (b *BinaryBuilder) getCmdArgs(ctx context.Context, binaryType string) ([][]
 	binaryTypeID, _ := manifest.SplitBinaryType(binaryType)
 
 	switch binaryTypeID {
+	case "javascript/v8":
+		fmt.Printf("`javascript/v8` binary type found... \n")
+		return [][]string{}, nil
```
maoueh, line 204: What would be the build command? Created issue to track that.

package.json
```diff
@@ -0,0 +1 @@
+{}
```
maoueh, line 1: What are those and `package-lock.json` at top-level of repository? Let's remove.

wasm/v8/helpers.go
```diff
@@ -0,0 +1,51 @@
+package v8
```
maoueh: (Comment about the filename)

`helpers.go` is a bit generic, use some more specific so it doesn't become and melting pot of everything. `bytes_helpers.go` is probably better.

wasm/v8/helpers.go
```diff
@@ -0,0 +1,51 @@
...
+		js = append(js, fmt.Sprint(val)...)
+	}
+	js = append(js, ']')
+	return string(js)
+}
+
+// InjectUint8Array returns JavaScript code that assigns a Uint8Array to globalThis
+func InjectUint8Array(varName string, data []byte) string {
+	return fmt.Sprintf("globalThis.%s = new Uint8Array(%s);", varName, ByteArrayToJSArray(data))
+}
```
maoueh: How much do we still need this now?

wasm/v8/helpers.go
```diff
@@ -0,0 +1,51 @@
...
+
+	for index := range length {
+		elem, _ := obj.GetIdx(uint32(index))
+		if !elem.IsNumber() {
+			return nil, fmt.Errorf("Uint8Array[%d] is not a number", index)
+		}
+		bytes[index] = byte(elem.Integer())
+	}
+	return bytes, nil
+}
```
maoueh: Same

wasm/v8/factory.go
```diff
@@ -0,0 +1,22 @@
+package v8
+
+import (
+	"context"
+
+	"github.com/streamingfast/substreams/wasm"
+	"rogchap.com/v8go"
+)
+
+type V8ModuleFactory struct{}
```
maoueh, line 10: Is this used, does not look like it.

wasm/v8/module.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (mod *V8Module) NewInstance(context.Context) (wasm.Instance, error) {
+	return NewV8Instance(mod.iso)
+}
+
+func (mod *V8Module) ExecuteNewCall(
+	ctx context.Context,
+	call *wasm.Call,
+	cachedInstance wasm.Instance,
+	_ []wasm.Argument,
```
maoueh, line 29: Why those are unused?

wasm/v8/module.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Used to inject input, entry buffer
+	var input []byte
+	for _, val := range argValues {
+		input = val
+		break
+	}
+	if input == nil {
+		input = []byte{}
+	}
```
maoueh, line 41: If there is multiple values, this would pick a random one from the map, as iteration order is not guaranteed. Seems to be missing multiple arguments support, I created an issue for that.

wasm/v8/module.go
```diff
@@ -0,0 +1,128 @@
...
+		scripts := []struct{ code, name string }{
+			{polyfillCode, "polyfill.js"},
+			{string(mod.code), "bundle.js"},
+		}
+		for _, s := range scripts {
+			if err := runJS(inst, s.code, s.name); err != nil {
+				inst.Close(ctx)
+				return nil, err
+			}
+		}
```
maoueh, line 59: Benchmark what is faster, added an issue for this.

wasm/v8/module.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if input == nil {
+		input = []byte{}
+	}
+
+	inst := getInstance(mod.iso, cachedInstance)
+
+	// Inject input before running scripts (needed to run main)
+	if err := runJS(inst, InjectUint8Array("input", input), "inject_input.js"); err != nil {
+		inst.Close(ctx)
```
maoueh: This is repeated over, could we have used a `defer` to ensure it's not missed?

wasm/v8/module.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	// call the actual js function
+	if err := callMain(inst); err != nil {
+		inst.Close(ctx)
+		return nil, err
+	}
+
+	outBytes, err := getOutput(inst)
+	_ = inst.ctx.Global().Set("output", v8go.Null(inst.ctx.Isolate()))
```
maoueh, line 69: Turn into function, created an issue for this.

wasm/v8/module.go
```diff
@@ -0,0 +1,128 @@
...
+	v8, _ := NewV8Instance(iso)
+	return v8
+}
+
+func runJS(inst *V8Instance, code, filename string) error {
+	if _, err := inst.ctx.RunScript(code, filename); err != nil {
+		return fmt.Errorf("executing %s: %w", filename, err)
+	}
+	return nil
+}
```
maoueh, line 98: Feel like a not that useful function, just call the right thing in-place instead IMO.

go.mod
```diff
@@ -206,11 +206,15 @@ require (
...
 
 retract v1.0.2 // Published at wrong tag.
 
 replace (
 	github.com/jhump/protoreflect => github.com/streamingfast/protoreflect v0.0.0-20231205191344-4b629d20ce8d
 	github.com/yourbasic/graph v0.0.0-20210606180040-8ecfec1c2869 => github.com/streamingfast/graph v0.0.0-20220329181048-a5710712d873
+	rogchap.com/v8go => github.com/streamingfast/v8go v0.0.0-20250530143322-fe2ca23f76b3
 )
+
+// replace rogchap.com/v8go => /home/ulysse/Documents/SF/v8go
```
maoueh, line 220: Remove.

wasm/v8/module.go
```diff
@@ -0,0 +1,191 @@
...
+		input = val
+		break
+	}
+	if input == nil {
+		input = []byte{}
+	}
+
+	inst := getInstance(mod.iso, cachedInstance)
+	inputVal, _ := v8go.NewValue(inst.ctx.Isolate(), input)
+	_ = inst.ctx.Global().Set("input", inputVal)
```
maoueh, line 45: There should be no `input` global variable, added an issue for this.

wasm/v8/module.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// Runs all scripts (will be changed depending on files needed), probably going to merge all that are needed. This if makes sure we load our needed scripts ONLY on the first call
+	if cachedInstance == nil {
+		_ = injectStoreFunction(inst.ctx, call)
```
maoueh, line 49: Why output unsued?

wasm/v8/module.go
```diff
@@ -0,0 +1,191 @@
...
+
+	return v8val.Uint8Array(), nil
+}
+
+func injectStoreFunction(ctx *v8go.Context, call *wasm.Call) error {
+	iso := ctx.Isolate()
+
+	storeSetFunc := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
+		if len(info.Args()) != 3 {
+			panic("__store_set expects 3 arguments")
```
maoueh, line 173: Let's ensure panics are properly trapped, added an issue for this.

wasm/v8/runtime/buf.gen.yaml
```diff
@@ -0,0 +1,5 @@
+version: v1
```
maoueh, line 1: Why do need package.json + buf stuff in `runtime`? Are they use for the benchmark?

wasm/v8/module.go
```diff
@@ -0,0 +1,191 @@
...
+	"context"
+	_ "embed"
+	"fmt"
+
+	"github.com/streamingfast/substreams/wasm"
+	"rogchap.com/v8go"
+)
+
+//go:embed runtime/polyfill.bundle.js
+var polyfillCode string
```
maoueh, line 13: Use a `[]byte` here, even if the file is plain text, a file should always be represented as a `[]byte`, can be done while doing the benchmarking of `runJS` concatenation.

wasm/v8/module.go
```diff
@@ -0,0 +1,191 @@
...
+
+func callHandlers(inst *V8Instance, call *wasm.Call) error {
+	handlerName := call.ModuleName
+
+	// Check what type of handler this is
+	checkTypeCode := fmt.Sprintf(`getHandlerType("%s")`, handlerName)
+	handlerTypeVal, err := inst.ctx.RunScript(checkTypeCode, "check_handler_type.js")
+	if err != nil {
+		return fmt.Errorf("failed to check handler type: %w", err)
+	}
```
maoueh, line 108: I don't understand, what is this exactly? Technically, we should know already what Substreams module we are executing so it's should be possible to query from somewhere.

wasm/v8/module.go
```diff
@@ -0,0 +1,191 @@
...
+func callMapHandler(inst *V8Instance, handlerName string) error {
+	// Call the map handler directly, input is already set as global
+	code := fmt.Sprintf(`executeMapHandler("%s", input)`, handlerName)
+	result, err := inst.ctx.RunScript(code, "execute_map_handler.js")
+	if err != nil {
+		return fmt.Errorf("failed to execute map handler '%s': %w", handlerName, err)
+	}
+
+	return inst.ctx.Global().Set("output", result)
+}
```
maoueh, line 135: We should call the class' instance directly, created an issue for that.

maoueh (changes requested):

PR #625, opened 2025-05-09
Feature/faster deleteprefix

sduchesneau:
- every deletePrefix caches the given prefix in a map.
- Next deletePrefix on the SAME prefix will be a noop.
- Any write to the store's map will look cause a lookup to the cache, invalidating any prefix matched by that entry.

This addresses the scenario where a module will constantly try to delete stores that begin by the prefix of the "previous month" or similar.

### DIFF
--- a/pipeline/cache/engine.go
+++ b/pipeline/cache/engine.go
@@ -29,7 +29,7 @@ type Engine struct {
 	blockType         string
 	reversibleBuffers map[uint64]*execout.Buffer // block num to modules' outputs for that given block
 	execOutputWriters map[string]*execout.Writer // moduleName => writer (single file)
-	existingExecOuts  map[string]*execout.File
+	existingExecOuts  map[string]*execout.File   // on Tier2 requests, this contains any existing outputs that we could load from disk, skipping module execution
 	indexWriters      map[string]*index.Writer
 
 	logger *zap.Logger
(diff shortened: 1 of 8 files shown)


pipeline/cache/engine.go
```diff
@@ -88,6 +88,9 @@ func (e *Engine) HandleFinal(clock *pbsubstreams.Clock) error {
 	}
 
 	delete(e.reversibleBuffers, clock.Number)
+	for _, existingExecOut := range e.existingExecOuts { // delete mapper outputs from previous blocks to free up memory ASAP. The existingExecOuts are only used on tier2
+		delete(existingExecOut.Kv, clock.Id)
```
maoueh: Can you explain in the comment here why it's safe to delete them now?

storage/store/delta.go
```diff
@@ -21,6 +22,12 @@ func (b *baseStore) ApplyDelta(delta *pbsubstreams.StoreDelta) {
 	keySize := uint64(len(delta.Key))
 	switch delta.Operation {
 	case pbsubstreams.StoreDelta_UPDATE:
+		for k := range b.recentlyDeletedPrefixes {
+			if strings.HasPrefix(delta.Key, k) {
+				delete(b.recentlyDeletedPrefixes, k) // we added a key matching this prefix, it is not considered deleted anymore
+			}
+		}
+
```
maoueh: This is repeated many times, it feels like maybe you should have a struct `DeletedPrefixesCache` or something like that and a single method would clean.

storage/store/value_delete.go
```diff
@@ -17,6 +17,13 @@ func (b *baseStore) DeletePrefix(ord uint64, prefix string) {
 }
 
 func (b *baseStore) deletePrefix(ord uint64, prefix string) {
+	if _, ok := b.recentlyDeletedPrefixes[prefix]; ok {
+		return
+	}
+	if len(b.recentlyDeletedPrefixes) > 100 {
+		b.recentlyDeletedPrefixes = make(map[string]struct{}) // keep this under reasonable size
+	}
```
maoueh: Maybe we should use a simple cache library, I used some that were very easy to configure, might not be required.

maoueh (changes requested):

storage/store/value_delete.go
sduchesneau: I checked, but won't go that route because it is not needed (more complex) and there are many choices, all with their specific performance caveats..

PR #628, opened 2025-05-16
refactor: use the built-in max/min to simplify the code

pengqiseven:
In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.

### DIFF
--- a/storage/index/configs.go
+++ b/storage/index/configs.go
@@ -16,10 +16,7 @@ type Configs struct {
 func NewConfigs(baseObjectStore dstore.Store, allRequestedModules []*pbsubstreams.Module, moduleHashes map[string]string, firstStreamableBlock uint64, logger *zap.Logger) (*Configs, error) {
 	out := make(map[string]*Config)
 	for _, mod := range allRequestedModules {
-		initialBlock := mod.InitialBlock
-		if initialBlock < firstStreamableBlock {
-			initialBlock = firstStreamableBlock
-		}
+		initialBlock := max(mod.InitialBlock, firstStreamableBlock)
 		conf, err := NewConfig(
 			mod.Name,
 			initialBlock,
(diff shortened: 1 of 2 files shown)


storage/execout/streamproto/protostream_test.go
```diff
@@ -128,10 +128,8 @@ func TestEncode(t *testing.T) {
 	if !bytes.Equal(encodedFull, encodedStream) {
 		t.Logf("Encoding methods produced different results")
 		// You might want to examine the first few bytes to see where they differ
-		compareLength := min(len(encodedFull), len(encodedStream))
-		if compareLength > 20 {
-			compareLength = 20 // Limit to first 20 bytes for readability
-		}
```
maoueh: Revert that one, don't feel the nested `min` increase readability here.

storage/execout/configs.go
```diff
@@ -20,10 +20,7 @@ func NewConfigs(baseObjectStore dstore.Store, allRequestedModules []*pbsubstream
 	out := make(map[string]*Config)
 	for _, mod := range allRequestedModules {
 
-		initialBlock := mod.InitialBlock
-		if initialBlock < firstStreamableBlock {
-			initialBlock = firstStreamableBlock
-		}
+		initialBlock := max(mod.InitialBlock, firstStreamableBlock)
```
maoueh: Let's revert `mod.InitialBlock` and `firstStreamableBlock` in the three cases
```suggestion
		initialBlock := max(firstStreamableBlock, mod.InitialBlock)
```

maoueh (changes requested):

storage/execout/streamproto/protostream_test.go
pengqiseven: Thanks for your advice. Reverted.

storage/execout/configs.go
pengqiseven: reverted.

PR #631, opened 2025-06-04
Remove input global variable

UlysseCorbeil:


go.mod
```diff
@@ -215,7 +215,7 @@ retract v1.0.2 // Published at wrong tag.
 replace (
 	github.com/jhump/protoreflect => github.com/streamingfast/protoreflect v0.0.0-20231205191344-4b629d20ce8d
 	github.com/yourbasic/graph v0.0.0-20210606180040-8ecfec1c2869 => github.com/streamingfast/graph v0.0.0-20220329181048-a5710712d873
-	rogchap.com/v8go => github.com/streamingfast/v8go v0.0.0-20250530143322-fe2ca23f76b3
+	rogchap.com/v8go => github.com/streamingfast/v8go v0.0.0-20250604153749-c36ab3bbf86d
 )
 
-// replace rogchap.com/v8go => /home/ulysse/Documents/SF/v8go
+//replace rogchap.com/v8go => /home/ulysse/Documents/SF/v8go
```
maoueh: To be removed.

wasm/v8/module.go
```diff
@@ -97,54 +93,81 @@ func runJS(inst *V8Instance, code, filename string) error {
...
+	switch handlerTypeVal.String() {
 	case "map":
-		return callMapHandler(inst, handlerName)
+		return callMapHandler(inst, handlerName, input)
 	case "store":
-		return callStoreHandler(inst, call, handlerName)
+		return callStoreHandler(inst, handlerName, input)
 	default:
-		return fmt.Errorf("unknown handler type: %s", handlerType)
+		return fmt.Errorf("unknown handler type: %s", handlerTypeVal.String())
```
maoueh:
```suggestion
		return fmt.Errorf("unknown handler type: %s", handlerTypeVal)
```
`%s` calls `.String()` automatically if defined.

maoueh: `handlerTypeVal.String()`, could prove better to cache it to a variable, unsure how costly it is from an execution standpoint.

wasm/v8/module.go
```diff
@@ -97,54 +93,81 @@ func runJS(inst *V8Instance, code, filename string) error {
...
+	outputVal, _ := v8go.NewUint8Array(inst.ctx, input)
+
+	// Define the store interface (JS object) that maps to the Go __store_set
+	storeInterfaceCode := `({ set: function(ordinal, key, value) {
+								__store_set(ordinal, key, value);
+							}})`
+	storeIface, err := inst.ctx.RunScript(storeInterfaceCode, "store_iface.js")
+	if err != nil {
+		return fmt.Errorf("could not create store interface: %w", err)
+	}
```
maoueh, line 188: Ideally this would be removed at some point or all moved into the `polyfill.js` bundle, should be execute once for each instance, no need to execute on each call.

maoueh (changes requested):

wasm/v8/module.go
maoueh: Actually, this seems like it should go in the `@substreams/sdk` library

UlysseCorbeil: This will be added with the stores PR

PR #635, opened 2025-06-05
First integration test

UlysseCorbeil:
https://linear.app/streamingfast/issue/BLO-461/first-integration-tests


package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
maoueh: Maybe it would be good to actually ignore the file altogether :)

test/testdata/js_substreams/src/lib.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export default class Substreams {
+	@substreams.handlers.map([BlockSchema], MapResultSchema)
+	test_map(blk: Block) {
+		return create(MapResultSchema, {
+			blockNumber: blk.number,
+			blockHash: blk.id,
+		});
+	}
+}
```
maoueh, line 13: This looks great in the end, really happy about the simple case

maoueh (approved):

PR #637, opened 2025-06-10
feature/lazyloadexecouts

sduchesneau:
Rework the execout File read/write:

This reduces the RAM usage necessary to read and stream data to the user on tier1,
as well as to read the existing execouts on tier2 jobs (in multi-stage scenario)

The cached execouts need to be rewritten to take advantage of this, since their data is currently not ordered:
the system will automatically load and rewrite existing execout when they are used.

Code changes include:

* new FileReader / FileWriter that "read as you go" or "write as you go"
* No more 'KV' map attached to the File
* Split the IndexWriter away from its dependencies on execoutMappers.
* Clock distributor now also reads "as you go", using a small "one-block-cache"

Removed env var and behaviors:
* SUBSTREAMS_DISABLE_PRELOAD_EXEC_FILES (no more preloading, it was mostly useful because reading full file+unmarshal was necessary when streaming...)
* SUBSTREAMS_OUTPUT_SIZE_LIMIT_PER_SEGMENT (not a RAM issue anymore)


orchestrator/execout/execout_walker.go
```diff
@@ -115,10 +105,14 @@ func computeNewWait(previousWait time.Duration, storeIsLocal bool) time.Duration
...
+func (r *Walker) sendItems(get func() (*pboutput.Item, error)) error {
+	for {
+		item, err := get()
+		if err != nil {
+			return err
+		}
 		if item == nil {
-			continue // why would that happen?!
+			return nil // why would that happen?!
 		}
```
maoueh: If the contract is for get to return either `<non-nil>, nil` or `nil, <error>`, we should panic here. 

Also, you should look at new Go iterator interface, we could implement it and consuming for greater conformance to how iterator should be defined now in Go.

service/tier2.go
```diff
@@ -414,13 +413,7 @@ func (s *Tier2Service) processRange(ctx context.Context, request *pbssinternal.P
...
 			case <-time.After(10 * time.Second): // 10 sec
 				if time.Since(now) > 300*time.Second { // 300 secs
-					size := 0
-					for _, file := range executionPlan.ExistingExecOuts {
-						for _, v := range file.Kv {
-							size += len(v.Payload)
-						}
-					}
-					logger.Info("request active for for a long time", zap.Duration("duration", time.Since(now)), zap.Int("execout_size_mib", size/1024/1024))
+					logger.Info("request active for for a long time", zap.Duration("duration", time.Since(now)))
```
maoueh:
```suggestion
					logger.Info("request active for a long time", zap.Duration("duration", time.Since(now)))
```

service/tier2.go
```diff
@@ -499,20 +492,18 @@ excludable:
...
-			}
-		}
 
-		sortedClocksDistributor := sortClocksDistributor(clocksDistributor)
 		ctx, span := reqctx.WithSpan(ctx, "substreams/tier2/pipeline/mapper_stream")
-		for _, clock := range sortedClocksDistributor {
-			if clock.Number < startBlock || clock.Number >= stopBlock {
-				panic("reading from mapper, block was out of range") // we don't want to have this case undetected
+
+		distributor := execout.NewClockDistributor(executionPlan.ExistingExecOuts, startBlock, stopBlock)
```
maoueh, line 498: Same here about Go iterators ..., would normally allow for `for (x, y) in range distributor { ... }`

storage/execout/file.go
```diff
@@ -2,213 +2,332 @@ package execout
...
+	*File
+	reader       io.ReadCloser
+	lastReadItem *pboutput.Item
+	complete     bool
+}
 
-	writingFile   *io.PipeWriter
-	writeError    chan error
-	deletedBefore *uint64
+func NewFileReader(ctx context.Context, store dstore.Store, logger *zap.Logger, rng *block.Range, moduleName string) (FileReader, error) {
```
maoueh:
```suggestion
func OpenFileReader(ctx context.Context, store dstore.Store, logger *zap.Logger, rng *block.Range, moduleName string) (FileReader, error) {
```

The name would be more closely aligned with `os.OpenFile` for example. Otherwise, maybe separating the `New` and the `Open` on the struct would be better.

storage/execout/file.go
```diff
@@ -2,213 +2,332 @@ package execout
...
+	done               chan struct{}
 }
-func (c *File) Filename() string {
-	return computeDBinFilename(c.Range.StartBlock, c.Range.ExclusiveEndBlock)
+
+func (fw *fileWriter) Range() *block.Range {
+	return fw.File.Range
+}
+
+type ClockDistributor struct {
```
maoueh, line 73: I like the name

storage/execout/file.go
```diff
@@ -2,213 +2,332 @@ package execout
...
-				Id:        item.BlockId,
-				Timestamp: item.Timestamp,
+func (cd *ClockDistributor) Next(ctx context.Context) (*pbsubstreams.Clock, error) {
+	for i := cd.nextClockNumber; i < cd.stopBlock; i++ {
+		for name, execOut := range cd.execOuts {
+			for {
+				lastRead, ok := cd.execOutsLastclock[name]
+				if ok && lastRead >= i {
+					break
+				}
```
maoueh, line 100: Does this. happen on network with skip block?

storage/execout/file.go
```diff
@@ -2,213 +2,332 @@ package execout
...
+		}
+
 	}
+	return nil, io.EOF
 }
 
-func (c *File) SetItem(clock *pbsubstreams.Clock, data []byte) error {
-	c.Lock()
-	defer c.Unlock()
+func NewFileWriter(ctx context.Context, store dstore.Store, logger *zap.Logger, rng *block.Range, moduleName string) FileWriter {
```
maoueh, line 146: That is also more than a `New` IMO

storage/execout/file.go
```diff
@@ -2,213 +2,332 @@ package execout
...
+	sort.Slice(items, func(i, j int) bool {
+		return items[i].BlockNum < items[j].BlockNum
+	})
 
-		outputData := &pboutput.Map{}
-		if err = outputData.UnmarshalFast(bytes); err != nil {
-			return fmt.Errorf("unmarshalling file %s: %w", filename, err)
+	store.SetOverwrite(true)
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel() // closes the WriteAsYouGo
```
maoueh, line 354: the `WriteAsYouGo` don't make sense anymore as it's been replaced with something else no?

storage/execout/file.go
```diff
@@ -2,213 +2,332 @@ package execout
...
+	if err != nil {
+		return err
+	}
+	if !ordered {
+		if err := rewriteAsOrdered(ctx, r, readBytes, fr.store, fr.ModuleName(), fr.Filename(), fr.Range, fr.logger); err != nil {
+			return err
+		}
+		if err := r.Close(); err != nil {
+			return err
+		}
```
maoueh: Does `rewriteAsOrdered` will write back to the same file that was read? I think then that `r.Close()` should be closed before attempting the write.

On local disk for example, I think the write would always fail file would be busy.

storage/execout/file.go
```diff
@@ -2,213 +2,332 @@ package execout
...
+		ordered, readBytes, err = streamproto.ReadOrderedBool(r)
+		if err != nil {
+			return err
+		}
+		if !ordered {
+			return fmt.Errorf("internal error: could not rewrite outputs as ordered")
+		}
+		fr.reader = r
```
maoueh: Do we really need this, seems a bit defensive, if we can't trust what we just wrote ...

Or could there be a race here with overlapping processes?

storage/execout/file.go
```diff
@@ -2,213 +2,332 @@ package execout
...
-		c.Kv = outputData.Kv
-
-		c.logger.Debug("outputs data loaded", zap.Int("output_count", len(c.Kv)), zap.Stringer("block_range", c.Range))
-		return nil
-	})
-	if err == nil {
-		c.loaded = true
 	}
-	return err
+	return newFile.Close()
```
maoueh: There is an exit path in the for/loop above, a defer should be used to close the file.

storage/execout/file_test.go
```diff
@@ -1,36 +0,0 @@
-package execout
```
maoueh, line 1: Not relevant anymore?

tools/decode.go
```diff
@@ -370,19 +370,18 @@ func searchOutputsModule(
...
-		if err == dstore.ErrNotFound {
-			return fmt.Errorf("can't find cache at block %d storeURL %q", startBlock, moduleStore.BaseURL().String())
-		}
-
-		return fmt.Errorf("loading cache %s file %s : %w", moduleStore.BaseURL(), outputCache.String(), err)
+	outputCache, err := modStore.NewFileReader(ctx, rng)
+	if err != nil {
+		return fmt.Errorf("can't create file reader for module %q: %w", module.Name, err)
 	}
+	fmt.Println("filename:", outputCache.Filename())
```
maoueh: Leftover

tools/decode.go
```diff
@@ -421,19 +420,18 @@ func searchOutputsModuleKvOps(
...
```
maoueh: Leftover

maoueh (changes requested):

storage/execout/file_test.go
sduchesneau: no it was just testing the clock extractor thingy

PR #639, opened 2025-06-11
test bench for polyfill and concat

UlysseCorbeil:


wasm/bench/bench_test.go
```diff
@@ -44,28 +44,28 @@ func BenchmarkExecution(b *testing.B) {
 		// {"bare", "map_noop", args(wasm.NewParamsInput("")), []int{0}},
 
 		// Decode proto only decode and returns the block.number as the output (to ensure the block is not elided at compile time)
-		{"decode_proto_only", "map_decode_proto_only", blockInputFile(b, "testdata/ethereum_mainnet_block_16021772.binpb"), []int{0}},
+		// {"decode_proto_only", "map_decode_proto_only", blockInputFile(b, "testdata/ethereum_mainnet_block_16021772.binpb"), []int{0}},
```
maoueh: Can't we use some test name to avoid comment/decommenting each time we need to play in some benchmark

wasm/bench/v8_bench_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	ctx := newV8Instance(b)
+	defer ctx.Isolate().Dispose()
+
+	_, _ = ctx.RunScript(string(combined), "polyfill_and_bundle.js")
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
```
maoueh: Let's use new Golang 1.24 `.Loop` feature instead of `N` https://go.dev/blog/testing-b-loop

wasm/bench/v8_bench_test.go
```diff
@@ -0,0 +1,64 @@
...
+func BenchmarkV8_RunJS_Concat(b *testing.B) {
+	poly := mustRead(b, "testdata/polyfill.bundle.js")
+	bund := mustRead(b, "testdata/bundle.js")
+
+	combined := append(poly, bund...)
+
+	ctx := newV8Instance(b)
+	defer ctx.Isolate().Dispose()
+
+	_, _ = ctx.RunScript(string(combined), "polyfill_and_bundle.js")
```
maoueh: Why do we need to run it once?

wasm/bench/testdata/polyfill.bundle.js
```diff
@@ -0,0 +1,3312 @@
+// This is free and unencumbered software released into the public domain.
```
maoueh: Can't we just create that in the benchmark setup code? I don't think there is necessity to commit that since we can easily concatenate

wasm/bench/v8_bench_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	ctx := newV8Instance(b)
+	defer ctx.Isolate().Dispose()
+
+	if _, err := ctx.RunScript(string(poly), "polyfill.js"); err != nil {
+		b.Fatalf("polyfill init error: %v", err)
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
```
maoueh: This reset all timer and metrics losing all the work done prior it, don't think your number are actually meaningful.

wasm/v8/benchmark.md
```diff
@@ -61,3 +65,19 @@ for 10k blocks this translates to (approximately) 4.73 minutes for the rust vers
 This time is different when running on docker since we're in production mode and have access to the cache. When using the cache the difference between the two is negligible but when un-cached the difference is very noticeable.
 
 On a clean instance you can see a 50-70% slower execution on the JS side than the rust-wasm execution, no cache, for a 1000 blocs. This trend is also true for a number of blocks n that converge to 10k blocks.
+
+### Polyfill + Bundle Loading Benchmark
+
+Benchmark comparing two strategies for loading JS scripts on a fresh context:
```
maoueh, line 71: Didn't though of telling you that the Go has a tool `benchstat` that can easily compare two benchmarks and note the difference: https://betterstack.com/community/guides/scaling-go/golang-benchmarking/#comparing-benchmarks-with-benchstat

maoueh (changes requested):

PR #642, opened 2025-06-11
clock intrinsics

UlysseCorbeil:


test/testdata/js_substreams/substreams.yaml
```diff
@@ -22,3 +22,10 @@ modules:
           - source: sf.substreams.v1.test.Block
       output:
           type: proto:sf.substreams.v1.test.MapResult
+    - name: test_map_clock
+      kind: map
+      initialBlock: 10
+      inputs:
+          - source: sf.substreams.v1.Clock
```
maoueh: The point is actually to avoid this, see other comment above.

test/testdata/js_substreams/src/lib.ts
```diff
@@ -10,4 +11,12 @@ export default class Substreams {
 			blockHash: blk.id,
 		});
 	}
+	@substreams.handlers.map([ClockSchema], MapResultSchema)
```
maoueh: Why is `ClockSchema` define in the handler? The `map` has no input, which is actually a bit weird, don't think it would be should be valid actually now in JavaScript.

I see it's probably because you define `Clock` input in the `substreams.yaml`. Let's use still a `BlockSchema` as input as we want people to not define Clock and leverage the intrinsics actually.

wasm/v8/module.go
```diff
@@ -46,7 +48,10 @@ func (mod *V8Module) ExecuteNewCall(
 
 	// Runs all scripts (will be changed depending on files needed), probably going to merge all that are needed. This if makes sure we load our needed scripts ONLY on the first call
 	if cachedInstance == nil {
+
 		_ = injectStoreFunction(inst.ctx, call)
+		_ = injectClockFunction(inst.ctx, call.Clock)
```
maoueh: Either check the return value or omit it completely. Also, the full code to create an instance should be move to its own function I think, at least soon as we are going to add more and more stuff.

wasm/v8/module.go
```diff
@@ -165,6 +170,28 @@ func getOutput(inst *V8Instance) ([]byte, error) {
...
 }
 
+// Injects __clock into runtime
+func injectClockFunction(ctx *v8go.Context, clock *pbsubstreams.Clock) error {
+	iso := ctx.Isolate()
+
+	data, err := proto.Marshal(clock)
+	if err != nil {
+		return fmt.Errorf("marshal clock: %w", err)
+	}
```
maoueh: Wrong, you are marshalling when injecting the function, which means the clock will remain constant, check multiple block output in your tests, it will probably fail in that case.

The clock must be provided at time of calling not for the full cached instance.

It's possible we might need to:
- Do some magic to have a single bounding of `__clock` but which implementation knows where to get the active clock for the current call.
- Re-bound the `__clock` function for each block

test/js_substreams_test.go
```diff
@@ -75,3 +78,65 @@ func Test_JSRuntime_SimpleMap(t *testing.T) {
...
+
+	jsHex := hex.EncodeToString([]byte(moduleName))
+	outputFile := fmt.Sprintf("%s/outputs/0000000010-0000000020.output", jsHex)
+
+	out, err := readOutputFile(ctx, filepath.Join(testTempDir, "test.store", "tag"), outputFile)
+	require.NoError(t, err)
+
+	var result pbsubstreamstest.MapResult
+
+	item, found := out.Kv["block-10"]
```
maoueh: All blocks should be validated, see other comment

test/js_substreams_test.go
```diff
@@ -75,3 +78,65 @@ func Test_JSRuntime_SimpleMap(t *testing.T) {
 
 	assertFiles(t, testTempDir, false, withZST([]string{outputRel})...)
 }
+
+func Test_JSRuntime_ClockIntrinsic(t *testing.T) {
+	manifest.TestUseSimpleHash = true
```
maoueh: Fully repeated from above, unless there is a compelling reason to start our own re-usable functions for testing JS, I would like we go back to the cases approach under the current integration test framework.

test/testdata/js_substreams/src/lib.ts
```diff
@@ -10,4 +11,12 @@ export default class Substreams {
...
+	test_map_clock() {
+		const clock = getClock();
```
maoueh, line 15: I can't see, but have you documented the `getClock` function in the SDK? Use proper JavaDoc there to document the behavior and what it does.

wasm/v8/module.go
```diff
@@ -165,6 +170,28 @@ func getOutput(inst *V8Instance) ([]byte, error) {
...
+
+	clockVal, err := v8go.NewUint8Array(ctx, data)
+	if err != nil {
+		return fmt.Errorf("clock Uint8array : %w", err)
+	}
+
+	clockFunc := v8go.NewFunctionTemplate(iso, func(info *v8go.FunctionCallbackInfo) *v8go.Value {
```
maoueh: Like maybe here, `FunctionCallbackInfo` could be use to retrieve some "dynamic" data, data which would be set in the `v8go` instance/module just before invoking it.

maoueh (changes requested):

PR #651, opened 2025-07-08
add version notation and tests

UlysseCorbeil:


manifest/manifest.go
```diff
@@ -66,6 +67,11 @@ func (m *Manifest) resolvePath(path string) string {
 		return path
 	}
 
+	parts := strings.Split(path, "@")
+	if len(parts) == 2 && (semver.IsValid(parts[1]) || parts[1] == "latest") {
+		return path
+	}
+
```
maoueh, line 74: Do we have this code elsewhere already?

maoueh (changes requested):
Some questions to ensure consistency.

manifest/manifest.go
UlysseCorbeil: Kind of, in reader.go :

https://github.com/streamingfast/substreams/blob/ac08f85f0e55dd1aeea2828f8fd803b293b90873/manifest/reader.go#L517

but not for this specific part

PR #661, opened 2025-07-24
Foundational Store Substreams integration

UlysseCorbeil:


UlysseCorbeil:
Warning: A git pull messed up the files changed, so theres a bunch of them that shouldn't be associated with this PR.

UlysseCorbeil:
Sorry about the commit history, I had to rebase so github didnt show hundreds of "changed" files coming from a merge

client/foundational/client.go
```diff
@@ -0,0 +1,64 @@
...
+		grpc.WithBlock(),
+		grpc.WithTimeout(5 * time.Second),
+	}
+
+	conn, err := dgrpc.NewExternalClientConn(rawEndpoint, opts...)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return &Store{rpc: pbstore.NewStoreKVClient(conn)}, conn.Close, nil
```
maoueh: Should we start without the `KV` prefix, even though we have a backing KV store, it could be something else, maybe it would be better to think in terms of the generic Foundational Store API than in a particular implementation.

client/foundational/client.go
```diff
@@ -0,0 +1,64 @@
...
+func (s *Store) GetAll(ctx context.Context, block uint64, keys [][]byte) (*pbstore.GetAllResponse, error) {
+	resp, err := s.rpc.GetAll(ctx, &pbstore.GetAllRequest{
+		Keys:        keys,
+		BlockNumber: block,
+		OmitDeleted: true,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return resp, nil
```
maoueh: I don't see the block hash reaching the server here, it's missing...

manifest/manifest.go
```diff
@@ -206,9 +211,110 @@ func (i *Input) parse() error {
...
+func (i *Input) parseFoundationalStore() error {
+	// Handle package notation
+	if strings.Contains(i.FoundationalStore, "@") {
+		_, _, _, err := parseFoundationalStoreIdentifier(i.FoundationalStore)
+		return err
+	}
+
+	if strings.HasPrefix(i.FoundationalStore, "grpc://") {
+		return nil
+	}
```
maoueh: I'm even wondering if we should continue to support this, my feeling tells me we should remove it. Let's remove it, more review made me realize it's really not here we will handle that.

manifest/manifest.go
```diff
@@ -206,9 +211,110 @@ func (i *Input) parse() error {
...
+	} else if version == "vlatest" {
+		errors = append(errors, fmt.Errorf("version %q is not supported for foundational-store inputs, use specific version", version))
+	}
+
+	if len(errors) > 0 {
+		return packageName, version, isShortcut, fmt.Errorf("invalid foundational-store identifier %q: %w", input, joinErrors(errors))
+	}
+
+	return packageName, version, isShortcut, nil
+}
```
maoueh: We already have this code to parse such package, let's re-use it.

manifest/manifest.go
```diff
@@ -206,9 +211,110 @@ func (i *Input) parse() error {
...
+	if len(errors) == 1 {
+		return errors[0]
+	}
+
+	var messages []string
+	for _, err := range errors {
+		messages = append(messages, err.Error())
+	}
+	return fmt.Errorf("%s", strings.Join(messages, "; "))
+}
```
maoueh: `errors.Join` is available now.

manifest/manifest.go
```diff
@@ -409,6 +515,20 @@ func (m *Module) setInputsToProto(pbModule *pbsubstreams.Module) error {
...
+					FoundationalStore: &pbsubstreams.Module_FoundationalStore{
+						// Endpoint: input.FoundationalStore,
+						Endpoint: resolvedEndpoint,
+					},
+				},
+			}
+			pbModule.Inputs = append(pbModule.Inputs, pbInput)
+			continue
+		}
+
```
maoueh: The resolution needs to happen at runtime and not at compile time. At compile time we must keep the input as specified by the user.

proto/sf/substreams/v1/modules.proto
```diff
@@ -83,12 +83,17 @@ message Module {
     string output_type = 1;
   }
 
+  message FoundationalStore {
+    string endpoint = 1;
+  }
```
maoueh: Fully document all this structure. Also, let's not use `endpoint` since we will not store endpoint in the final package. 

`id` or `identifier` or something in this vein is more approriate.

client/foundational/client.go
```diff
@@ -0,0 +1,64 @@
...
+
+	opts := []grpc.DialOption{
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithUnaryInterceptor(otelgrpc.UnaryClientInterceptor()),
+		grpc.WithStreamInterceptor(otelgrpc.StreamClientInterceptor()),
+		grpc.WithBlock(),
+		grpc.WithTimeout(5 * time.Second),
+	}
+
+	conn, err := dgrpc.NewExternalClientConn(rawEndpoint, opts...)
```
maoueh: Those should be InternalClient actually.

manifest/manifest.go
```diff
@@ -206,9 +211,110 @@ func (i *Input) parse() error {
...
+	case "local", "dev", "development":
+		return "grpc://localhost:50051"
+	case "staging", "stage":
+		return "grpc://foundational-store-staging:10016"
+	case "production", "prod":
+		return "grpc://foundational-store:10016"
+	}
+
+	return "grpc://foundational-store:10016"
+}
```
maoueh: To be re-thinked, needs to happen at runtime and must supports multiple mappings.

manifest/manifest_test.go
```diff
@@ -1,6 +1,7 @@
 package manifest
```
maoueh, line 1: Let's ensure also that foundational-store is correctly taken into consideration for hash calculation and that two exactly same manifest diverging only by a change in the version of foundational-store input correctly yield a new hash.

manifest/signature.go
```diff
@@ -173,6 +175,8 @@ func inputValue(input *pbsubstreams.Module_Input) (string, error) {
 		return "", nil // this is accounted for in the `AncestorOf()` tree
 	case *pbsubstreams.Module_Input_Map_:
 		return "", nil // this is accounted for in the `AncestorOf()` tree
+	case *pbsubstreams.Module_Input_FoundationalStore:
+		return input.GetFoundationalStore().GetEndpoint(), nil
```
maoueh: Think this correctly handles the hash of the input.

pipeline/exec/baseexec.go
```diff
@@ -150,7 +153,7 @@ func (e *BaseExecutor) wasmCall(outputGetter execout.ExecutionOutputGetter, canS
 			}
 			return nil, fmt.Errorf("block %d: module %q: general wasm execution failed: %w, %w", clock.Number, e.moduleName, err, ctxErr)
 		}
-		return nil, fmt.Errorf("block %d: module %q: general wasm execution failed: %w: %s", clock.Number, e.moduleName, wasm.ErrWasmDeterministicExec, err)
+		return nil, fmt.Errorf("block %d: module %q: general wasm execution failed 3: %w: %s", clock.Number, e.moduleName, wasm.ErrWasmDeterministicExec, err)
```
maoueh: Leftover the `3`?

pipeline/pipeline.go
```diff
@@ -774,6 +793,14 @@ func (p *Pipeline) cleanUpModuleExecutors(ctx context.Context) error {
...
 		}
 	}
+	// tear down any foundational store gRPC clients
+	for endpoint, closers := range p.foundationalClosers {
+		for _, closeFn := range closers {
+			if err := closeFn(); err != nil {
+				logger.Warn("closing foundational store client", zap.String("endpoint", endpoint), zap.Error(err))
+			}
+		}
+	}
```
maoueh, line 812: I'm actually wondering if we should actually keep a client more globally, and not tied to a particular pipeline run. This way, we avoid having to establish and tear down connections altogether.

Maybe an optimization for later.

pipeline/process_block.go
```diff
@@ -455,7 +455,13 @@ func (p *Pipeline) execute(ctx context.Context, executor exec.ModuleExecutor, ex
...
 				}
 			}
-			panic(r) // send other panics up one level
+			if s, ok := r.(string); ok {
+				e := fmt.Errorf("wasmtime error: %s", s)
+				p.execoutStorage.ConfigMap[executorName].WriteDeterministicError(ctx, execOutput.Clock().Number, e)
+				out.err = e
+				return
+			}
+			panic(fmt.Errorf("unknown error: %s", r))
```
maoueh: Would like to understand better this change. Why it would be `wasmtime` here actually? We have more than one vm processor.

Also, when recovering from panic, just checking for `r.(string)` is usually not enough, you need to check for `error` too.

test/integration_test.go
```diff
@@ -58,7 +58,7 @@ func TestForkHandling(t *testing.T) {
 			exclusiveEnd: 7,
 			production:   true,
 			forkBlockRefs: []*ForkBlockRef{
-				{blockRef: bstream.NewBlockRef("1a", 1), previousID: "0a", libBlockRef: bstream.NewBlockRef("0a", 0)},
+				{blockRef: bstream.NewBlockRef("1a", 1), previousID: "34a", libBlockRef: bstream.NewBlockRef("0a", 0)},
```
maoueh: Failure?

test/tier2_integration_test.go
```diff
@@ -57,6 +57,9 @@ func TestTier2Call(t *testing.T) {
 	mapHybridInputBlock70 := hex.EncodeToString([]byte("map_hybrid_input_block_70"))
 	setSumStoreInit0 := hex.EncodeToString([]byte("set_sum_store_init_0"))
 
+	// foundationMapHash := hex.EncodeToString([]byte("foundation_map"))
+	// foundationStoreHash := hex.EncodeToString([]byte("foundational_store"))
+
```
maoueh: Remove

wasm/wazero/state_hostmod.go
```diff
@@ -372,6 +374,92 @@ var StateFuncs = []funcs{
...
+				return
+			}
+
+			// Serialize response
+			respData, err := proto.Marshal(resp)
+			if err != nil {
+				call.ReturnError(fmt.Errorf("failed to marshal GetAllResponse: %w", err))
+				stack[0] = 0
+				return
+			}
```
maoueh: Would be great to leave a note to explore if we could actually make a `DoFoundationalStoreGetAll` but instead of receiving a `GetAllResponse` and then doing `proto.Marshal`, if we could receive simply the `[]byte` without any re-decoding done.

maoueh: Applies to other similar implementation

wasm/wazero/state_hostmod.go
```diff
@@ -372,6 +374,92 @@ var StateFuncs = []funcs{
...
+
```
maoueh: Maybe we should create an helper for those as we will also need in JavaScript!

maoueh (changes requested):

pipeline/process_block.go
UlysseCorbeil: There was a silent panic here, it just failed without crashing iirc so I added an error we can see and if it ever reaches that panic, means something went very wrong

PR #662, opened 2025-07-28
add undo manager to cancel streams doing an external call (ex: eth_call) when a block gets forked out

sduchesneau:


pipeline/exec/baseexec.go
```diff
@@ -121,10 +121,17 @@ func (e *BaseExecutor) wasmCall(outputGetter execout.ExecutionOutputGetter, canS
...
+	ctx := e.ctx
 	if sharedCache.Cachable(clock.Number) {
-		err = sharedCache.Execute(e.ctx, e.wasmModule, e.moduleHash, call, e.wasmArguments, argValues)
+		err = sharedCache.Execute(ctx, e.wasmModule, e.moduleHash, call, e.wasmArguments, argValues, undoManager)
 	} else {
-		inst, err = e.wasmModule.ExecuteNewCall(e.ctx, call, e.cachedInstance, e.wasmArguments, argValues)
+		if undoManager != nil {
+			var cancel func()
+			// note: a canceled context only has an effect inside 'external calls', not within the wasm module executor itself, so the cancelation does not affect all connected streams
+			ctx, cancel = undoManager.Subscribe(ctx, clock.Id)
```
maoueh: Also here `unsubscribe`?

pipeline/exec/undo_manager.go
```diff
@@ -0,0 +1,142 @@
...
+
+var GlobalUndoManager *UndoManager
+var ErrBlockUndo = errors.New("block undone")
+var MaxUndoneBlocksSize = 100
+
+var globalIDCounter atomic.Uint64
+
+func nextUniqueID() uint64 {
+	return globalIDCounter.Add(1)
+}
```
maoueh: Tie it to `UndoManager` struct instead of being a global, id needs to be unique per manager anyway.

pipeline/exec/undo_manager.go
```diff
@@ -0,0 +1,142 @@
...
+func (u *UndoManager) Contains(id string) bool {
+	u.Lock()
+	_, ok := u.previousUndoneBlocks[id]
+	u.Unlock()
+	return ok
+}
+
+// Subscribe returns a context that will get canceled if the block with the given ID gets previousUndoneBlocks
+// returns an unsubscribe function that must be called to prevent leaks
+func (u *UndoManager) Subscribe(ctx context.Context, id string) (context.Context, func()) {
```
maoueh:
```suggestion
func (u *UndoManager) Subscribe(ctx context.Context, blockID string) (context.Context, func()) {
```

Using `id` collapses with `watcher` `id` and confused me a lot.

pipeline/exec/undo_manager.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// UndoManager gives cancelable contexts as subscriptions to specific block hashes. if that blocks gets an UNDO signal from a reorg, it will cancel that context.
+type UndoManager struct {
+	sync.Mutex
+	activeSubscriptions  map[string][]watcher
+	previousUndoneBlocks map[string]uint64
+}
+
+func NewUndoManager(cancelDelay time.Duration) *UndoManager {
```
maoueh: Unused argument

pipeline/exec/undo_manager.go
```diff
@@ -0,0 +1,142 @@
...
+	return &UndoManager{
+		activeSubscriptions:  make(map[string][]watcher),
```
maoueh, line 34: Seems that nobody is clearing `activeSubscriptions`, think there is a memory leak around it where it would forever grow (with empty arrays as values since subscription themself got deleted from the array, but not the key).

maoueh (changes requested):

PR #667, opened 2025-08-05
add atomic op to sink save

UlysseCorbeil:


sink/cursor.go
```diff
@@ -21,10 +22,31 @@ func ReadCursor(filename string) (*Cursor, error) {
...
 	if filename == "" || cursor == nil {
 		return nil
 	}
-	return os.WriteFile(filename, []byte(cursor.String()), 0644)
+
+	dir := filepath.Dir(filename)
+	tempFile, err := os.CreateTemp(dir, ".cursor_*")
+	if err != nil {
+		return err
+	}
```
maoueh, line 35:
```go
	tempFile, err := os.CreateTemp(nil, ".cursor_*")
	if err != nil {
		return err
	}
```

No need for a temp directory, can pass `nil`

sink/cursor.go
```diff
@@ -21,10 +22,31 @@ func ReadCursor(filename string) (*Cursor, error) {
...
+		os.Remove(tempPath)
+		return err
+	}
+
+	if err = tempFile.Close(); err != nil {
+		os.Remove(tempPath)
+		return err
+	}
+
+	return os.Rename(tempPath, filename)
```
maoueh, line 51: If rename fails, doesn't it leave the temporary file on disk?

Probably a `defer os.Remove(tempPath)` would work in all cases, even if renamed, think it's going to cause no problem.

sink/cursor.go
UlysseCorbeil: Wouldn't passing nil to CreateTemp break atomicity since it would use the default temp dir which might be on a different filesystem than the target file ?

PR #677, opened 2025-08-19
refactor: update worker pool to use WorkerBroker interface

colindickson:
## Summary
This PR refactors the orchestrator worker-pool to improve separation of concerns, testability, and concurrency safety by introducing a small `WorkerBroker` abstraction and a gRPC adapter. The pool core is decoupled from transport details; deterministic unit tests are added for `GlobalWorkerPool`.

### Key changes
- Added: `orchestrator/work/broker.go` — `WorkerBroker` interface + `GRPCWorkerBroker` adapter.
- Added: `orchestrator/work/broker_test.go` — adapter forwarding tests.
- Added: `orchestrator/work/globalworkerpool_test.go` — deterministic tests for borrow/return/keep-alive.
- Modified: `orchestrator/work/globalworkerpool.go` — depends on `WorkerBroker`, safe map iteration, atomic ramp-up/counters.
- Modified: `orchestrator/work/worker.go` — keep-alive plumbing uses `WorkerBroker`.
- Modified: `orchestrator/work/factory.go` — wraps gRPC client via `NewGRPCWorkerBroker`.

No public API outside `orchestrator` was changed. Runtime wiring remains unchanged because the factory adapts the concrete gRPC client to the new interface.

---

## Motivation
This change isolates responsibilities, reduces coupling to generated gRPC types, and makes the core pool logic easier to test and extend.

SOLID scores (subjective, 0–100) — before → after

- Single Responsibility (S)
  - Before: 50 → After: 85
  - Rationale: Transport and pool responsibilities were mixed; now transport is an adapter and the pool focuses on lifecycle and policy.

- Open/Closed (O)
  - Before: 40 → After: 75
  - Rationale: New broker interface enables extension (alternate transports, fakes) without modifying core pool code.

- Liskov Substitution (L)
  - Before: 70 → After: 85
  - Rationale: Small interface makes substituting fakes/adapters straightforward.

- Interface Segregation (I)
  - Before: 60 → After: 85
  - Rationale: Consumers depend only on methods they need instead of a large generated client surface.

- Dependency Inversion (D)
  - Before: 30 → After: 90
  - Rationale: High-level pool depends on an abstraction (`WorkerBroker`) rather than concrete gRPC types.

Average: Before ≈ 50 / After ≈ 84

---

## Problems addressed
- Testability: Decouples transport from pool via the `WorkerBroker` interface so tests can inject lightweight fakes or mocks instead of real gRPC clients. This makes it practical to write deterministic unit tests for pool behaviour 
- Tight transport coupling: Pool no longer depends on generated gRPC client types; adapter patterns centralize transport code.
- Concurrent map iteration during shutdown: safe iteration pattern (collect keys under lock then iterate) prevents races.
- Ambiguous ramp-up flags: atomic counters replace boolean flags for safe concurrent access.
- Extensibility: Borrow/return policies and transport implementations can be swapped with minimal changes.

---

## Roadmap / context
This PR is intentionally small and focused: it extracts a narrow transport abstraction and improves testability for the worker-pool. Consider this a prelude to other future `orchestrator` refactors and simplifications.


maoueh:
If rebased, would be good to be included.

PR #681, opened 2025-08-26
add cache to speed up buf gen and fix BSR limit

UlysseCorbeil:


manifest/protobuf.go
```diff
@@ -136,23 +185,39 @@ func loadDescriptorSets(pkg *pbsubstreams.Package, manif *Manifest) ([]*desc.Fil
 			"https://buf.build",
 		)
 
-		request := connect.NewRequest(&reflectv1beta1.GetFileDescriptorSetRequest{
-			Module:  descriptor.Module,
-			Symbols: descriptor.Symbols,
-			Version: descriptor.Version,
-		})
+		// Try to load from cache first
+		cacheKey := generateCacheKey(descriptor.Module, descriptor.Version, descriptor.Symbols)
```
maoueh: What is the value of `Version` here, in substreams manifest, I think we use nothing as the version query:

```
protobuf:
  descriptorSets:
    - module: buf.build/streamingfast/firehose-stellar
    - module: buf.build/protocolbuffers/wellknowntypes
```

This would be problematic, also true for "symlink" like `buf.build/protocolbuffers/wellknowntypes@main`, if you cache the `main` version, you will never update it.

manifest/protobuf.go
```diff
@@ -22,6 +24,53 @@ import (
 	"google.golang.org/protobuf/types/descriptorpb"
 )
 
+func getCacheDir() (string, error) {
+	cacheDir := filepath.Join("/tmp", "substreams-buf.cache")
+	return cacheDir, os.MkdirAll(cacheDir, 0755)
+}
```
maoueh: `os.UserConfigDir()` seems more appropriate for this task, more permanent also, we would want the cache to live for a while, some system delete `/tmp` often, OSX does it for me from time to time.

manifest/protobuf.go
```diff
@@ -22,6 +24,53 @@ import (
 	"google.golang.org/protobuf/types/descriptorpb"
 )
 
+func getCacheDir() (string, error) {
+	cacheDir := filepath.Join("/tmp", "substreams-buf.cache")
+	return cacheDir, os.MkdirAll(cacheDir, 0755)
```
maoueh: I would separate the MkdirAll call and the value retrieval concerns. Doing a Mkdir system call to the OS each time you `loadFromCache` is wasted, there is a need to create the folder only once.

Which points me that a `struct` named `<Something>Cache` would solve this "state" problem.

manifest/protobuf.go
```diff
@@ -22,6 +24,53 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	data, err := proto.Marshal(fds)
+	if err != nil {
+		return err
+	}
+
+	cacheFile := filepath.Join(cacheDir, cacheKey+".pb")
```
maoueh:
```suggestion
	cacheFile := filepath.Join(cacheDir, cacheKey+".file_descriptor.binpb")
```

The extra file_descriptor describes a bit the content, `binpb` is the Buf recognized extension for binary protobuf.

manifest/protobuf_test.go
```diff
@@ -349,3 +350,255 @@ func TestReaderOptions_ProtoPathAndDescriptorSet(t *testing.T) {
...
+
+	if len(keys) != 4 {
+		t.Errorf("Expected 4 unique keys, got %d", len(keys))
+	}
+}
+
+func TestCacheSaveAndLoad(t *testing.T) {
+	// Create a temporary cache directory for testing
+	tempCacheDir := filepath.Join("/tmp", "substreams-buf-test.cache")
+	defer os.RemoveAll(tempCacheDir)
```
maoueh:
```suggestion
	tempCacheDir := t.TempDir()
```

Will auto-cleanup, nothing to do more.

maoueh: The cache dir needs of course to be controllable.

manifest/protobuf_test.go
```diff
@@ -349,3 +350,255 @@ func TestReaderOptions_ProtoPathAndDescriptorSet(t *testing.T) {
...
+			},
+		},
+	}
+
+	cacheKey := "test_cache_key_1234567890abcdef1234567890abcdef12345678"
+
+	// Test save to cache
+	err := saveToCache(cacheKey, testFds)
+	if err != nil {
+		t.Fatalf("Failed to save to cache: %v", err)
```
maoueh: Tell your Claude to use `require` and `assert` library for tests, they render errors more nicely and compact the test cases.

maoueh: True for all tests here.

manifest/protobuf_test.go
```diff
@@ -349,3 +350,255 @@ func TestReaderOptions_ProtoPathAndDescriptorSet(t *testing.T) {
...
+		
+		if cachedFile.GetPackage() != originalFile.GetPackage() {
+			t.Errorf("Package name mismatch at index %d: expected %s, got %s", 
+				i, originalFile.GetPackage(), cachedFile.GetPackage())
+		}
+
+		if len(cachedFile.MessageType) != len(originalFile.MessageType) {
+			t.Errorf("Message type count mismatch at index %d: expected %d, got %d", 
+				i, len(originalFile.MessageType), len(cachedFile.MessageType))
+		}
```
maoueh: `proto.Equal` needs to be used to determine protobuf equality. Usually we couple that with a `require.EqualExported` (not sure of the exact name but check equality only on exported fields of a struct) which check correct equality rules and print a diff if it's not good:

```
if !proto.Equal(a, b) {
  require.EqualExported(t, a, b) // Fails and print diff
}
```

maoueh (changes requested):
Good overall idea, needs to flesh out some details to be effective.

UlysseCorbeil:
Alright so I modified what was talked about and the cache works but theres still an issue, since caching dosen't happen on the first build we still hit the BSR rate limiting issue and since our descriptorSets module are version-less eg:

```yaml
descriptorSets:
    - module: buf.build/streamingfast/substreams-sink-sql
    - module: buf.build/streamingfast/substreams-foundational-store
    - module: buf.build/streamingfast/firehose-solana
```

thats what the time.sleep fixed (temporarly) so we either need to add versions to our modules or add a small sleep for first time builds always and then on subsequent build the cache will kick in

manifest/protobuf.go
```diff
@@ -19,10 +20,66 @@ import (
...
+	return semver.IsValid(version)
+}
+
+func (c *DescriptorCache) generateCacheKey(module, version string, symbols []string) string {
+	h := sha256.New()
+	h.Write([]byte(module))
+	h.Write([]byte(version))
+	for _, symbol := range symbols {
+		h.Write([]byte(symbol))
+	}
```
maoueh, line 69: What are the symbols here we write? I think it would be good to document them if needed to compute the cache key as I have no idea what they are.

manifest/protobuf.go
```diff
@@ -19,10 +20,66 @@ import (
...
+	return hex.EncodeToString(h.Sum(nil))
+}
+
+func (c *DescriptorCache) load(cacheKey string) (*descriptorpb.FileDescriptorSet, error) {
+	cacheFile := filepath.Join(c.cacheDir, cacheKey+".file_descriptor.binpb")
```
maoueh: This should be shared between save and load

manifest/protobuf.go
```diff
@@ -103,6 +160,12 @@ func loadDescriptorSets(pkg *pbsubstreams.Package, manif *Manifest) ([]*desc.Fil
 		seen[*file.Name] = true
 	}
 
+	// Initialize cache once for all descriptor sets
+	cache, err := newDescriptorCache()
+	if err != nil {
+		return nil, fmt.Errorf("initializing descriptor cache: %w", err)
+	}
```
maoueh: We should never fail initializing the cache, we should fallback to either an in-memory cache or no caching, a warning is fine though.

manifest/protobuf.go
```diff
@@ -137,23 +200,61 @@ func loadDescriptorSets(pkg *pbsubstreams.Package, manif *Manifest) ([]*desc.Fil
...
-			Version: descriptor.Version,
-		})
+		var fileDescriptorSetResponse *descriptorpb.FileDescriptorSet
 
-		authToken := os.Getenv("BUFBUILD_AUTH_TOKEN")
-		if authToken != "" {
-			request.Header().Set("Authorization", "Bearer "+authToken)
-		}
+		// Don't use cache for mutable versions or empty versions
+		if !cache.shouldCache(descriptor.Version) {
```
maoueh:
```suggestion
		if !cache.isDeterministicVersion(descriptor.Version) {
```

maoueh: I think the inverse flow would read better and reduce the duplication you have now between the two if branches.

```
if cache.isDeterministicVersion(descriptor.Version) {
  desc, err := cache.load(...)
  if err == nil {
    return desc
  } else {
    // Log warning about cache retrieval failure, I'm even fine if it's just logged in DEBUG
  }
}

// Fetch & save from here as either it was not cached or some error happened.

File: manifest/protobuf.go [hunk 12]
```diff
@@ -137,23 +200,61 @@ func loadDescriptorSets(pkg *pbsubstreams.Package, manif *Manifest) ([]*desc.Fil
...
+					Symbols: descriptor.Symbols,
+					Version: descriptor.Version,
+				})
+
+				authToken := os.Getenv("BUFBUILD_AUTH_TOKEN")
+				if authToken != "" {
+					request.Header().Set("Authorization", "Bearer "+authToken)
+				}
+
+				fileDescriptorSet, err := client.GetFileDescriptorSet(context.Background(), request)
```
maoueh: You should almost never use `context.Background`, in your `loadLocalProtobufs`, it should receive a `context` up until a caller has some context that make sense.

At the CLI level, we often use `cmd.Context()`

manifest/protobuf.go
```diff
@@ -137,23 +200,61 @@ func loadDescriptorSets(pkg *pbsubstreams.Package, manif *Manifest) ([]*desc.Fil
...
+				if err != nil {
+					return nil, fmt.Errorf("getting file descriptor set for %s: %w", descriptor.Module, err)
+				}
+
+				fileDescriptorSetResponse = fileDescriptorSet.Msg.FileDescriptorSet
+
+				// Save to cache (ignore errors to not break the build)
+				_ = cache.save(cacheKey, fileDescriptorSetResponse)
```
maoueh: Always check for error, log as warning or debug

manifest/protobuf.go
```diff
@@ -165,8 +266,8 @@ func loadDescriptorSets(pkg *pbsubstreams.Package, manif *Manifest) ([]*desc.Fil
 			seen[fd.GetName()] = true
 			out = append(out, fd)
 		}
-		// TEMP FIX FOR TEST DEPLOYMENT
-		time.Sleep(1020 * time.Millisecond)
+		// TEMP FIX FOR TEST DEPLOYMENT (keeping this here still until we've decided)
+		// time.Sleep(1020 * time.Millisecond)
```
maoueh: Leftover, to be removed.

maoueh (changes requested):

maoueh (approved):
Just missing a Changelog entry and we are good to merge.

maoueh (approved):

PR #690, opened 2025-09-23
Feature/request v3

sduchesneau:
This allows sending the full spkg instead of just the modules in the substreams request.
The server can do the mangling of replacing the network and params with a provided value.


app/tier1.go
```diff
@@ -21,8 +21,7 @@ import (
 	"github.com/streamingfast/shutter"
 	"github.com/streamingfast/substreams/client"
 	"github.com/streamingfast/substreams/metrics"
-	"github.com/streamingfast/substreams/pb/sf/substreams/rpc/v2/pbsubstreamsrpcconnect"
-	ssconnect "github.com/streamingfast/substreams/pb/sf/substreams/rpc/v2/pbsubstreamsrpcconnect"
+	ssconnect "github.com/streamingfast/substreams/pb/sf/substreams/rpc/v2/pbsubstreamsrpcv2connect"
```
maoueh: Even if it's long, I think we should still use `pbsubstreamsrpcv2connect`, IMO.

client/client.go
```diff
@@ -88,7 +94,7 @@ func (c *SubstreamsClientConfig) MarshalLogObject(encoder zapcore.ObjectEncoder)
 
 type InternalClientFactory = func() (cli pbssinternal.SubstreamsClient, closeFunc func() error, callOpts []grpc.CallOption, headers Headers, err error)
 
-func NewSubstreamsClientConfig(endpoint string, authToken string, authType AuthType, insecure bool, plaintext bool, agent string) *SubstreamsClientConfig {
+func NewSubstreamsClientConfig(endpoint string, authToken string, authType AuthType, insecure bool, plaintext bool, agent string, forceV2 bool) *SubstreamsClientConfig {
```
maoueh: We should probably think about using a struct as this function parameters, we added some parameters over time and I think we probably will add more after this one, up to you to refactor or not.

This will now be a breaking change also for people consuming `substreams` as a `substreams-sink` library.

maoueh: Should we specify forced version instead of v3, like `forceVersion string` (or enum would be better).

client/client.go
```diff
@@ -278,8 +287,18 @@ func NewSubstreamsClient(config *SubstreamsClientConfig) (cli pbsubstreamsrpc.St
 		}
 	}
 
-	zlog.Debug("creating new client", zap.String("endpoint", endpoint))
-	cli = pbsubstreamsrpc.NewStreamClient(conn)
+	return
+}
+
+func NewSubstreamsClients(config *SubstreamsClientConfig) (cliV2 pbsubstreamsrpcv2.StreamClient, cliV3 pbsubstreamsrpcv3.StreamClient, closeFunc func() error, callOpts []grpc.CallOption, headers Headers, err error) {
```
maoueh: Should it be `NewSubstreamsClientConn(...) ...` and caller would be responsible of :

```
cliV2 := pbsubstreamsrpcv2.NewStreamClient(conn)
cliV3 := pbsubstreamsrpcv3.NewStreamClient(conn)
```

docs/release-notes/change-log.md
```diff
@@ -23,6 +23,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Server and Client
 
 * Added support for foundational-store (in wasmtime and wazero).
+* Added support for new 'sf.substreams.rpc.v3.Stream/Blocks' endpoint that sends the full '.spkg' data with params and network value, so the client does not need to do any mangling.
+  * This requires the substreams server to support it (under `/sf.substreams.rpc.v2.Stream/*` location).
```
maoueh: Hmmm `...rpc.v3`?

docs/release-notes/change-log.md
```diff
@@ -23,6 +23,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  * On the `run`, `gui` and `sink` commands, the `--force-v2` flag is available to keep the previous protocol; the `v2` endpoint will also be tried as fallback if the server responds with 404 or MethodNotAllowed.
```
maoueh: To be thought of in relation to maybe yet a future v4.

pb/sf/substreams/rpc/v2/substreams.go
```diff
@@ -1,4 +1,4 @@
-package pbsubstreamsrpc
+package pbsubstreamsrpcv2
```
maoueh, line 1: We should think about this rename if truly necessary, we could change only our internal usage, that would make the Protobuf fully the same as before.

pb/sf/substreams/rpc/v3/request.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return &pbsubstreamsrpcv2.Request{
+		StartBlockNum:                       r.StartBlockNum,
+		StartCursor:                         r.StartCursor,
+		StopBlockNum:                        r.StopBlockNum,
+		FinalBlocksOnly:                     r.FinalBlocksOnly,
+		ProductionMode:                      r.ProductionMode,
+		OutputModule:                        r.OutputModule,
+		Modules:                             r.Package.Modules,
+		DebugInitialStoreSnapshotForModules: r.DebugInitialStoreSnapshotForModules,
```
maoueh, line 27: We should probably take the time to remove this one entierly from v3, it think it has ben unsupported for a long time now.

pb/sf/substreams/rpc/v3/request.go
```diff
@@ -0,0 +1,92 @@
...
+func ParamsToMap(in []string) (map[string]string, error) {
+	if len(in) == 0 {
+		return nil, nil
+	}
+	out := make(map[string]string)
+	for _, p := range in {
+		kv := strings.SplitN(p, "=", 2)
+		if len(kv) != 2 {
+			return nil, fmt.Errorf("invalid parameter format: %s", p)
+		}
```
maoueh, line 44: There is `strings.Cut` now that exists to make that "clearer", in this particular case, it would be mostly equivalent.

proto/sf/substreams/rpc/v3/service.proto
```diff
@@ -0,0 +1,72 @@
...
+option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/rpc/v3;pbsubstreamsrpcv3";
+
+service Stream {
+  rpc Blocks(Request) returns (stream sf.substreams.rpc.v2.Response); // v3 request still sends v2 response
+}
+
+message Request {
+  int64 start_block_num = 1;
+  string start_cursor = 2;
+  uint64 stop_block_num = 3;
```
maoueh, line 34: Let's ensure we properly document all fields properly with the right amount of details, like precedence of `start_cursor` if both `start_block_num` and `start_cursor` are provided as well as how to follow chain's head.

proto/sf/substreams/rpc/v3/service.proto
```diff
@@ -0,0 +1,72 @@
...
+import "sf/substreams/v1/modules.proto";
+import "sf/substreams/v1/package.proto";
+
+option go_package = "github.com/streamingfast/substreams/pb/sf/substreams/rpc/v3;pbsubstreamsrpcv3";
+
+service Stream {
+  rpc Blocks(Request) returns (stream sf.substreams.rpc.v2.Response); // v3 request still sends v2 response
+}
+
+message Request {
```
maoueh, line 23: Documentation

proto/sf/substreams/rpc/v3/service.proto
```diff
@@ -0,0 +1,72 @@
...
+  uint64 limit_processed_blocks = 12;
+
+  // Available only in developer mode
+  // If set, the server will only send the output of these modules.
+  // If unset, all the outputs are sent.
+  // This allows the user to reduce the payload in dev mode, while still getting some extra debugging information
+  repeated string dev_output_modules = 13;
+
+  // Progress_messages_interval_ms is the interval between progress messages, in milliseconds (minimum: 500, default: start at 500ms and ramp up to 5000ms within 1min)
+  uint64 progress_messages_interval_ms = 14;
```
maoueh: Should we take the occassion to change the default?

tools/prometheus-exporter.go
```diff
@@ -50,6 +51,7 @@ func init() {
 	prometheusCmd.Flags().String("substreams-api-key-envvar", "SUBSTREAMS_API_KEY", "Name of variable containing Substreams Api Key")
 	prometheusCmd.Flags().BoolP("insecure", "k", false, "Skip certificate validation on GRPC connection")
 	prometheusCmd.Flags().BoolP("plaintext", "p", false, "Establish GRPC connection in plaintext")
+	prometheusCmd.Flags().Bool("force-v2", false, "Force using V2 API")
```
maoueh: This is missing details, unclear what this V2 API is about.

Side question, should we move `prometheus` to `substreams sink ...`? Maybe it's too specific to our needs though.

maoueh (changes requested):

PR #694, opened 2025-10-01
add short-hand notation for descriptorSets

UlysseCorbeil:


UlysseCorbeil:
We can now use the 3 differents versions:
```yaml
  descriptorSets:
    - module: buf.build/streamingfast/substreams-sink-sql
    - module: buf.build/streamingfast/substreams-foundational-store@v0.1.0
    - module: buf.build/streamingfast/firehose-solana@latest
```

manifest/manifest.go
```diff
@@ -530,3 +530,56 @@ func (m *Module) setOutputToProto(pbModule *pbsubstreams.Module) {
...
+
+	if likelyShortcut {
+		// User used @ notation
+		if validationErr != nil {
+			return fmt.Errorf("invalid descriptor set module notation %q: %w", module, validationErr)
+		}
+
+		if parsedVersion == "latest" {
+			return fmt.Errorf("descriptor set module %q: version 'latest' is not allowed, please specify a specific version or omit version entirely", module)
+		}
```
maoueh: That is interesting as we technically already accept them, but would make sense indeed to not accept literally `latest` and force people to use versionning.

manifest/manifest.go
```diff
@@ -530,3 +530,56 @@ func (m *Module) setOutputToProto(pbModule *pbsubstreams.Module) {
...
+
+		if version != "" {
+			return fmt.Errorf("descriptor set module %q: cannot specify version both in module field and version field", module)
+		}
+
+		// Expand to full buf.build path
+		b.Module = fmt.Sprintf("buf.build/streamingfast/%s", packageName)
```
maoueh: Not sure this is a good idea...

manifest/manifest.go
```diff
@@ -530,3 +530,56 @@ func (m *Module) setOutputToProto(pbModule *pbsubstreams.Module) {
...
+		b.Version = parsedVersion
+	} else if !strings.Contains(module, "/") {
+		// Short package name
+		b.Module = fmt.Sprintf("buf.build/streamingfast/%s", module)
```
maoueh: Same here.

maoueh (changes requested):
The thing I'm wondering though is that I don't see any changes at client level to actually make the query to buf Registry, don't wee need to pass something more now? 

Or maybe it was all good with `BufImport` struct and it was `""` before but now will be filled when using `@<version>`?

manifest/manifest.go
maoueh: What are the pros/cons in your view of enabling short notation for descriptorSets?

maoueh:
Something missing also here, update to Substreams reference doc about descriptorSets, documenting the new feature properly, maybe ask codegen to do it :)

manifest/manifest.go
UlysseCorbeil: Well it reduced verbosity, easier to maintain, gives us flexibility about versioning if we ever need to test with different versions, having this shorthand across the different pieces of a substreams module is also nice, since imports, descriptorSets and inputs like foundational-store also use this so theres a normalized way of doing this.

For cons I think maybe it hides complexity, since its not as explicit as a full on URL and its less obvious what the actual source is.

docs/new/references/substreams-components/manifests.md
```diff
@@ -145,6 +145,70 @@ Protobufs and modules are packaged together to help Substreams clients decode th
 
 [Learn more about Google Protocol Buffers](https://protobuf.dev/) in the official documentation provided by Google.
 
+#### `protobuf.descriptorSets`
```
maoueh, line 148: Beautiful

docs/new/references/substreams-components/manifests.md
```diff
@@ -145,6 +145,70 @@ Protobufs and modules are packaged together to help Substreams clients decode th
...
+    - module: buf.build/streamingfast/substreams-sink-sql@v0.1.0
+```
+{% endcode %}
+
+**Available Fields:**
+
+* `module` (required): The full path to the Buf module in the format `buf.build/organization/repository`. Can optionally include the version using `@version` notation.
+* `version` (optional): The semantic version of the module to import (e.g., `v0.1.0`, `v1.2.3`). Must be a valid [semver](https://semver.org/) format.
+* `symbols` (optional): An array of specific protobuf symbols to import from the descriptor set. If omitted, all types from the descriptor set are available.
+* `localPath` (optional): Local filesystem path where the descriptor set should be cached or stored.
```
maoueh, line 184: What is that, do we really have this feature right now? The usage of this seems weird.

docs/new/references/substreams-components/manifests.md
```diff
@@ -145,6 +145,70 @@ Protobufs and modules are packaged together to help Substreams clients decode th
...
+{% endcode %}
+
+**Version Validation Rules:**
+
+{% hint style="warning" %}
+**Important**: The following version rules are enforced:
+
+* The version **must** be a valid semantic version (e.g., `v1.0.0`, `v0.2.5`)
+* The version **cannot** be `latest` - you must specify an explicit version for reproducibility
+{% endhint %}
```
maoueh: Isn't this false for protobuf descriptorSets? I think it's only valid for `imports` no?

manifest/manifest.go
```diff
@@ -530,3 +531,44 @@ func (m *Module) setOutputToProto(pbModule *pbsubstreams.Module) {
...
+	version := strings.TrimSpace(b.Version)
+
+	modulePart, versionPart, hasAt := strings.Cut(module, "@")
+	if hasAt {
+		if version != "" {
+			return fmt.Errorf("descriptor set module %q: cannot specify version both in module field and version field", module)
+		}
+		if versionPart == "" || versionPart == "latest" {
+			return fmt.Errorf("descriptor set module %q: version 'latest' is not allowed, please specify a specific version or omit version entirely", module)
+		}
```
maoueh: This is true only when using `module: <name>@<version>` right?

maoueh (changes requested):
Small last stuff

docs/new/references/substreams-components/manifests.md
maoueh: Ok I checked, it's indeed used. My confusion is that the "source" struct is called `BufImport` while it's really not as for we have `protobuf.descriptorSets` and `localPath` represents a local file containing raw file descriptors.

So can you deprecated `BufImport` and rename to `DescriptorSetEntry`

```
// Deprecated: Use [DescriptorSetEntry] instead.
type BufImport = DescriptorSetEntry

type DescriptorSetEntry struct { 
...
}

...
```

manifest/manifest.go
```diff
@@ -80,19 +80,21 @@ type PackageMeta struct {
...
+	ExcludePaths   []string              `yaml:"excludePaths,omitempty"`
 }
 
-type BufImport struct {
+type DescriptorSetEntry struct {
 	LocalPath string   `yaml:"localPath,omitempty"`
 	Module    string   `yaml:"module"`
 	Version   string   `yaml:"version,omitempty"`
 	Symbols   []string `yaml:"symbols,omitempty"`
 }
```
maoueh, line 95: Missing deprecation notice

maoueh (changes requested):

maoueh (approved):

PR #698, opened 2025-10-16
add context cancellation upstream for calls

UlysseCorbeil:


pipeline/process_block.go
```diff
@@ -454,6 +454,9 @@ func (p *Pipeline) execute(ctx context.Context, executor exec.ModuleExecutor, ex
 	defer func() {
 		if r := recover(); r != nil {
 			if err, ok := r.(error); ok {
+				if errors.Is(err, context.Canceled) {
+					panic(context.Canceled)
+				}
```
maoueh: What is that? We panicked already, why changing the panic message in this case. This needs to be explained toroughly.

wasm/call.go
```diff
@@ -356,6 +366,14 @@ func (c *Call) DoFoundationalStoreGetAll(index uint32, block uint64, blockHash [
 	defer cancel()
 
 	for {
+		// Check if the call context was cancelled
+		select {
+		case <-c.ctx.Done():
+			zlog.Info("foundational store get_all cancelled due to upstream disconnect", zap.Uint64("block_number", block))
```
maoueh: Debug, this is expected, an info level is not needed here.

maoueh (changes requested):
The firehose-ethereum has WASM extension for `rpc_call` and `rpc_balance`, in there, it seems it receives `ctx`, I would like that you investigate how WASM extension deals with `ctx`, to ensure we have a similar behavior as here.

UlysseCorbeil:
Ok so the rpc extensions check ctx.Err() != nil after rpc errors and return the err without panicking. In our calls, errors go through ErrWasmDeterministicExec and panics so theres no way to return non deterministic error when our error gets wrapped in ErrWasmDeterministicExec. I added a custom error to handle this, our wasm intrinsics detect ctx.canceled and panic with this error, I had to keep the small logic in process_block because if we dont, it re-panics as a "uknown error" and prints the stack trace, but atleast now its a custom error so theres no danger from other sources "error-ing" into our panic check
Tell me what you think

PR #705, opened 2025-11-14
chore: fix a large number of spelling issues

tinyfoolish:
fix a large number of spelling issues


maoueh:
Those were actually old unused files and have been removed completely.

PR #713, opened 2025-12-02
Feature/latest fallback prise2

billettc:


reqctx/tier2request.go
```diff
@@ -18,10 +22,23 @@ type Tier2RequestParameters struct {
 }
 
 func WithTier2RequestParameters(ctx context.Context, parameters Tier2RequestParameters) context.Context {
-	return context.WithValue(ctx, tier2RequestParametersKeyKey, parameters)
+	fmt.Println("Setting tier2 request parameters", parameters)
```
sduchesneau: Get rid of this !!

sduchesneau (changes requested):
Remove println laying around (logs = $)
otherwise LGTM

PR #753, opened 2026-04-13
Feature/index optimisation

billettc:


app/tier2.go
```diff
@@ -61,6 +73,7 @@ func NewTier2(logger *zap.Logger, config *Tier2Config, modules *Tier2Modules) *T
 }
 
 func (a *Tier2App) Run() error {
+	ctx := context.Background()
```
sduchesneau, line 76: This context is only given to the moduleCache, but nothing cancels it or anything.
Why is it declared here ?

pipeline/cache/module.go
```diff
@@ -0,0 +1,70 @@
...
+	"context"
+	"sync"
+	"time"
+
+	"github.com/streamingfast/substreams/wasm"
+	"go.uber.org/zap"
+)
+
+type ModuleCache struct {
+	sync.Mutex
```
maoueh: `sync.RWMutex`

pipeline/cache/module.go
```diff
@@ -0,0 +1,70 @@
...
+	modules map[string]cacheEntry
+}
+
+type cacheEntry struct {
+	mod       wasm.Module
+	fetchedAt time.Time
+}
+
+func NewModuleCache(ctx context.Context, evictionInterval time.Duration, ttl time.Duration, logger *zap.Logger) *ModuleCache {
```
maoueh, line 22: What's the expected memory impact of the module cache?

pipeline/cache/module.go
```diff
@@ -0,0 +1,70 @@
...
+	"sync"
+	"time"
+
+	"github.com/streamingfast/substreams/wasm"
+	"go.uber.org/zap"
+)
+
+type ModuleCache struct {
+	sync.Mutex
+	modules map[string]cacheEntry
```
maoueh, line 14: Pretty sure a good concurrent map with bucket(s) will offer us a better performance since our module's hash would be fairly evenly distributed across the buckets.

pipeline/distributor/clock_distributor.go
```diff
@@ -0,0 +1,132 @@
...
+			clock, err := cd.Next(ctx)
+			if err == io.EOF {
+				return
+			}
+			if !yield(clock, err) {
+				return
+			}
+			if err != nil {
+				return
+			}
```
maoueh: Only the yielder should be responsible of exiting the loop normally. Someone using:

```
for clock, err := range distributor.Iter() {
  if err != nil { continue }
} 
```

Would silently stop the loop, I think a panic would be better here.

pipeline/pipeline.go
```diff
@@ -161,30 +166,32 @@ func New(
...
+		forkHandler:              NewForkHandler(),
+		blockStepMap:             make(map[bstream.StepType]uint64),
+		startTime:                time.Now(),
+		executionTimeout:         executionTimeout,
+		workerPoolFactory:        workerPoolFactory,
+		checkPendingShutdown:     checkPendingShutdown,
+		moduleNameToStage:        make(map[string]int),
+		outputBufferSize:         outputBufferSize,
+		supportBuffering:         supportBuffering,
+		StagedModuleExecutorsMap: make(map[string]*path.ExecutorPath),
```
maoueh, line 194: Why public?

service/tier2.go
```diff
@@ -143,9 +150,30 @@ func NewTier2(
 		opt(s)
 	}
 
+	if s.evictionInterval <= 0 {
+		s.evictionInterval = 60 * time.Second
+		s.logger.Warn("module cache evictionInterval was <= 0, reset to default", zap.Duration("value", s.evictionInterval))
```
maoueh: Maybe we should instead return an error if the module cache cannot be disabled.

service/tier2.go
```diff
@@ -143,9 +150,30 @@ func NewTier2(
...
+	}
+	if s.ttl <= 0 {
+		s.ttl = 10 * time.Minute
+		s.logger.Warn("module cache ttl was <= 0, reset to default", zap.Duration("value", s.ttl))
```
maoueh: Same here (and in tier1)

pipeline/distributor/clock_distributor.go
maoueh: I see it's an old code being moved, can be disregard

pipeline/pipeline.go
billettc: Use in service package ...

pipeline/pipeline.go
```diff
@@ -766,35 +773,62 @@ func (p *Pipeline) BuildModuleExecutors(ctx context.Context) error {
...
+	loadingModuleStart := time.Now()
+	moduleHashDuration := time.Duration(0)
 	for _, stage := range p.executionStages {
 		for _, layer := range stage {
 			for _, module := range layer {
 				code := reqModules.Binaries[module.BinaryIndex]
 				key := moduleKey{binaryIndex: module.BinaryIndex, binaryType: code.Type}
+
+				moduleHashStart := time.Now()
+				moduleHash := p.execGraph.ModuleHashes()[module.Name]
```
sduchesneau, line 793: seems weird to ask specifically for '[module.Name]' but then discard it.
These 2 lines would be more readable with:
`_ = p.execGraph.ModuleHashes() // memoize`

service/tier1.go
```diff
@@ -158,6 +159,7 @@ func getBlockTypeFromStreamFactory(sf *StreamFactory) (string, error) {
 	return strings.TrimPrefix(out, protoPkfPrefix), nil
 }
 func NewTier1(
+	ctx context.Context,
```
sduchesneau, line 162: same comment as for Tier2: passing a ctx here and not using it (dead code, misleading)
either the app has a context that it will cancel, either we don't pass down the context and we have a context.Background() used directly inside the cache.

sduchesneau (approved):
LGTM, just a few cosmetics suggested

PR #788, opened 2026-05-21
fix(ci): skip Docker login, push and Scout steps for fork PRs

maoueh:
The Docker build workflow already had a correct push condition but the registry login and Docker Scout steps ran unconditionally. For fork PRs the pushed image does not exist so Scout would fail; this change guards both steps with the same same-repo condition so fork PRs only build the image without pushing or scanning.


UlysseCorbeil (approved):
is .dev/todo supposed to be merged in ?

maoueh:
> is .dev/todo supposed to be merged in ?

Yes and no :) I'm not sure, removing for now and I'll continue to ponder if it's something we should or not.

PR #798, opened 2026-06-09
validate more fields in executed partial blocks to detect reorgs even when transaction hashes are the same

sduchesneau:
Problem:

If the state in reth flash-blocks executor is incorrect, it will only be detected
at the last flash-block, when the block hash is computed with its calculated state-root, 
validated against the next block's parent hash, then the block is recomputed.

Substreams detects that the block was replaced/recomputed base the list of transaction hashes,
it assumes that any change will come with different or disordered transactions.

If it does not (only the state is different, giving room to different events / state changes),
substreams will not detect the incorrect state and will not trigger a reorg.

Solution:

We now check more data on the transaction, including the full transaction receipt.
We changed the hashing algo for fnv which is better for small values (at least 2X, zero allocations, non-cryptographic)

### DIFF
--- a/pipeline/partials.go
+++ b/pipeline/partials.go
@@ -5,6 +5,7 @@ import (
 	"crypto/md5"
 	"errors"
 	"fmt"
+	"hash/fnv"
 
 	pbacme "github.com/streamingfast/dummy-blockchain/pb/sf/acme/type/v1"
 	pbeth "github.com/streamingfast/firehose-ethereum/types/pb/sf/ethereum/type/v2"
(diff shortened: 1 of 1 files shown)


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:d84062f</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:d84062f</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:a1ea31f02ed4e9034c8cd3eac8366685304ad02e5bb8e81a26bc0e0f29ba7977</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>117 MB</td></tr>
<tr><td>packages</td><td>353</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>c136481b8f4cd58cb213a22ee358ea1047bc963a338c61cdccb518e233409f86</code></li>

<li><code>noble</code></li>

<li><code>noble-20260509.1</code></li>
</ul>
</td></tr>
<tr><td>digest</td><td><code>sha256:023f8a753c22258c9fe2d0005a7d28258038da7d620e9f93e9ad78aa266f9f11</code></td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 24 low: 4 </td></tr>
</table>
</details>

<table></table>

maoueh (approved):

PR #821, opened 2026-07-06
Fix nil deref, swallowed cache read errors and log truncation cap

sduchesneau:
Fixes three confirmed bugs in the engine/cache/wasm paths.

## Fixes

1. **`pipeline/exec/sharedcache.go`** — `SharedCache.Execute` called `inst.Close(ctx)` before checking the error from `ExecuteNewCall`. The wazero runtime returns `(nil, err)` on instantiation or heap-write failure, so the nil interface dereference panicked in the tier1 shared-cache path and was recovered upstream as a generic "unknown error", masking the real cause. Now only closes a non-nil instance so the real error propagates.

2. **`pipeline/cache/engine.go`** — `Engine.NewBuffer` checked `if !ok { continue }` before `if err != nil`. `fileReader.Get` returns `(nil, false, err)` on read errors, so corrupt/failed execout cache reads were silently treated as "block absent", producing silent data gaps instead of a failure. The error check now comes first.

3. **`wasm/call.go`** — `ReachedLogsMaxByteCount` compared the cumulative `LogsByteCount` against `maxLogByteCount` (512 KiB per-message cap) instead of `maxTotalLogsByteCount`, truncating module logs 10x earlier than the truncation message itself claims. Fixed the comparison to use the total cap — and, since that alone would let logs grow up to the constant's previous value of 5 MiB (10x more output than what was actually ever enforced), lowered `maxTotalLogsByteCount` itself to 2 MiB so total log volume stays close to today's real-world behavior instead of jumping up.

## Tests

- `pipeline/exec/sharedcache_test.go` — fake `wasm.Module` whose `ExecuteNewCall` returns `(nil, err)`: `Execute` must return the error without panicking.
- `pipeline/cache/engine_test.go` — fake `execout.FileReader`: a read error must fail `NewBuffer`; absent blocks are still skipped; found payloads are still loaded into the buffer.
- `wasm/call_logs_test.go` — cumulative logs above 512 KiB but below 2 MiB must not be truncated; logs reaching 2 MiB must be truncated with the marker message.

All new tests fail on the unfixed code and pass after the fixes. Full `go test ./pipeline/... ./wasm/...` passes.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:df44e72</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:df44e72</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:a6f9ba37b8d411b6b73fa7b49ad054b6466f1a8b7467f6e2cf724b921d959164</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>110 MB</td></tr>
<tr><td>packages</td><td>361</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

maoueh (approved):

PR #822, opened 2026-07-06
Stream store quicksave lazily and parallelize across stores

sduchesneau:
## Problem

Store quicksave/quickload (used to save live store state on shutdown/disconnect and resume without reprocessing) had two scaling issues:

1. **Sequential across stores** — `Map.QuickSave`/`QuickLoad` looped one store at a time. Slow for pipelines with many stores, and the save path runs under a shutdown time budget.
2. **Eager serialization** — the streaming marshaler (`MarshalStream`) pre-computed the *entire* serialized store into memory (`precomputeChunks`) before the first `Read`. Despite the "streaming" plumbing, peak memory was still a full serialized copy alongside the live `kv` map.

## Change

- **Lazy marshaler**: `fastStreamingMarshaler` now encodes one KV entry (then delete-prefixes) at a time as `Read()` consumes it. Peak extra memory ≈ largest single entry + a values-free sorted-key slice, instead of the whole serialized store.
- **Bounded parallelism**: `QuickSave`/`QuickLoad` run up to 8 stores concurrently (`llerrgroup`). Streaming keeps per-store peak small, so the fan-out doesn't multiply whole-store buffers. Quickload memory-accounting (`AllocateFullKVSizeOrForceCancelRequest`) still runs serially after loads complete, since that handler isn't concurrency-safe.

## Bugfix (performance): tier2 job scheduled on a not-yet-final segment

In production mode, a request with a stop block below LIB but inside a segment that is not yet fully final (e.g. stop block 48285001 with head at 48285648, segment size 1000) computed the linear handoff as the *next* segment boundary (48286000). That scheduled a tier2 job for the whole segment, which hung waiting for merged blocks up to the segment end instead of streaming the few requested blocks.

`computeLinearHandoffBlockNum` now compares the segment boundary (`nextBoundary`) against LIB rather than the stop block: if the segment containing the stop block isn't fully final, handoff happens at the LIB boundary and tier1 streams the tail linearly. Segments fully below LIB are scheduled on tier2 as before.

## Compatibility

Wire format is **unchanged** — same `StoreData` protobuf bytes, same `.quicksave` filename. Keys are sorted (as the eager streaming path already did), so stream output stays deterministic. No migration; old and new binaries read each other's quicksave files. A truncated/partial file (e.g. crash mid-write) still fails `UnmarshalStream` cleanly → non-fatal fallback to reprocessing.

## Tests

New `vtproto_lazy_test.go` asserts the lazy path yields identical, wire-valid, deterministic bytes across read-buffer sizes (1 byte → 1 MiB, values larger than the buffer) and round-trips. Existing `storage/store/...` suite passes. New `g8_*` cases in `Test_computeLinaerHandoffBlockNum` cover the not-yet-final-segment handoff.

### DIFF
--- a/app/tier1.go
+++ b/app/tier1.go
@@ -176,7 +176,15 @@ func (a *Tier1App) Run() error {
 
 	var quickSaveStore dstore.Store
 	if a.config.QuickSaveStoreURL != "" {
-		quickSaveStore, err = dstore.NewStore(a.config.QuickSaveStoreURL, "zst", "zstd", true)
+		// Quicksave state is ephemeral (read once on resume, then discarded), so
+		// SUBSTREAMS_QUICKSAVE_COMPRESSION=none trades storage/egress for wall-clock
+		// by skipping zstd compression, which can dominate the save time for large
+		// stores on a fast same-region link.
+		quickSaveExt, quickSaveCompression := "zst", "zstd"
+		if os.Getenv("SUBSTREAMS_QUICKSAVE_COMPRESSION") == "none" {
+			quickSaveExt, quickSaveCompression = "", ""
+		}
+		quickSaveStore, err = dstore.NewStore(a.config.QuickSaveStoreURL, quickSaveExt, quickSaveCompression, true)
 		if err != nil {
 			return fmt.Errorf("failed setting up quickSave store from url %q: %w", a.config.QuickSaveStoreURL, err)
 		}
--- a/app/tier1.go
+++ b/app/tier1.go
@@ -176,15 +176,7 @@ func (a *Tier1App) Run() error {
 
 	var quickSaveStore dstore.Store
 	if a.config.QuickSaveStoreURL != "" {
-		// Quicksave state is ephemeral (read once on resume, then discarded), so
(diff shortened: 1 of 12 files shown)


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:5dc13b5</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:5dc13b5</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:8d603b107832e6251ab3560bbaa4f73b80b68f574a25684eccde10412ed18f7e</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>108 MB</td></tr>
<tr><td>packages</td><td>353</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

UlysseCorbeil (approved):
All good, fits fine, will have merge conflicts and will have to see what we do about MarshalStreamUnsorted probably but architecturally LGTM

PR #828, opened 2026-07-10
Feature/mmap store backend

UlysseCorbeil:
Adds an mmap (bbolt-backed) store backend as an alternative to the existing in-memory (Go heap) store, selectable via `--substreams-stores-backend=mmap`. The goal is to address production OOMs where large or concurrent FullKV stores exhaust a pod's heap: mmap keeps store data in a memory-mapped bbolt file so cold pages are reclaimable by the kernel under pressure, instead of pinning everything on the non-reclaimable Go heap. The in-memory backend stays the default; mmap is opt-in per deployment.

## Store backend management and configuration

- Added `StoresScratchSpace` and `StoresBackend` config fields to Tier1 and Tier2, wired through to the services.
- Updated test helpers and config constructors to accept the new backend parameters.

## Store correctness (mmap)

- The mmap backend is validated to produce byte-identical output to the in-memory backend, including through the delete-prefix and snapshot paths where bbolt's transaction-scoped value semantics require copying. Backed by mmap-specific regression tests (the store package previously only covered the memory backend).

## Resource cleanup and error handling

- Ensured store resources are properly closed to prevent leaks (new `Close` methods on `StoreModuleState` and `Stages`, plus cleanup in pipelines/orchestrators).
- Canceled loads no longer delete valid store files — a file is only treated as corrupt when the context is still live.

## Store load and metadata handling

- Store loads check the request context while streaming and abort promptly on cancellation; the post-load `SetMetadata` goroutine no longer captures the whole store and runs under a bounded timeout.
- Added `tryLoadFullKV` for context-aware, cache-independent loading.

## Metrics and observability

- New metrics for backend type, mmap file size, and mmap operations.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:eb77a7a</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:eb77a7a</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:35ac312452ff41be50c098518f3fa13377ea6a0ba1c606007a92571258a90e14</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>109 MB</td></tr>
<tr><td>packages</td><td>354</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

sduchesneau:
## Review — mmap (bbolt) store backend

Solid PR. The core aliasing/leak/cancel bugs are already caught and fixed well, and the `KVImpl` contract ("value valid only during callback, must not be retained") is documented. Both production retention sites (`merge.go`, `pipeline/snapshot.go`) copy correctly. Below is what the fixes missed or left inconsistent, plus production caveats.

### 1. Default backend contradicts the PR description — verify intent 🔴
`kv_impl_config.go:50,102` + the `runtimeconfig.go` comment: **mmap is the default** (`getKVImplTypeFromEnv` returns mmap unless env == `memory`; `NewKVImpl`'s default case is mmap). The description says *"in-memory backend stays the default; mmap is opt-in."* The code does the opposite. If the flag layer in firehose-core doesn't explicitly pass `memory`, **every deployment silently flips to mmap**. This is the single biggest production risk — confirm which is intended before merge.

### 2. `SetMetadata` goroutine in `FinalStoreMap` — the same bug the PR fixed elsewhere, not fixed here 🟠
`orchestrator/stage/stages.go:696-701` still does:
```go
go func() {
    err := fullKV.Store().SetMetadata(s.ctx, fullKV.Filename(), met)
    ...
}()
```
This captures the **whole `fullKV`** and runs on the raw request ctx — the exact anti-pattern commit `2ebbfa21` fixed in `pipeline.go:setupSubrequestStores` (narrow the capture to `metaStore/filename/name`, detach to a bounded 30s timeout). The tier1 twin was left untreated. Lower severity than the tier2 case (here `Close()` frees the data regardless, and the goroutine only touches objStore/filename), but it's inconsistent and should get the same treatment.

### 3. `keyCount` data race — untested, latent 🟠
`kv_impl_mmap.go:41` `keyCount int` is mutated (`++`/`--`) inside `Set`/`Delete`/`BatchSet`/`Load` while holding only `snapMu.RLock()` — the **shared** side. The struct comment explicitly says writes "may run concurrently with each other." Plain-int mutation + plain read in `KeyCount()` (line 557) under a shared lock = a data race under `-race` and torn counts. In practice a single store instance is single-writer (parallel jobs use separate instances), so it likely never fires — but nothing enforces it and there is no concurrency test. Either make `keyCount` atomic, or document/assert single-writer.

### 4. Memory vs mmap snapshot asymmetry — write-during-save hazard 🟡
mmap `Snapshot()` gates all writes via exclusive `snapMu` for the iterator's lifetime. Memory `Snapshot()` (`kv_impl_memory.go:156`) reads the **live map with no gate**. With the new *lazy* streaming marshal, the read window now spans the entire upload, not just `Save()`-time. If the "store frozen during save" invariant ever breaks, mmap blocks safely while memory gets a **concurrent-map-access fatal panic**. Fine today because save is synchronous on the processing goroutine — but the two backends now have different safety envelopes. Worth a comment at minimum.

### 5. Snapshot holds the exclusive lock for the full upload 🟡
`FullKV.Save()` acquires the snapshot lock, returns `fw`, and the caller uploads later under that lock. `Close()` also waits on it. A multi-GB save over slow object storage blocks all writes + `Close` on that store for the upload duration. Intended/documented, bounded by ctx cancellation of `Write` — but a stuck object-store PUT wedges the store until the deadline. Acceptable; flagging for ops awareness.

### Smaller / caveats
- **`newMmapKVImpl` / `newMmapKVImplNoSync`** (`:95`, `:110`) build the path from `os.Getpid()` only — two stores in one process collide on the same file. Only the `WithConfig` path (`CreateTemp`, unique) is used in prod; the PID ones look test/fallback-only. Latent footgun if ever promoted.
- **Silent mmap→memory fallback** (`kv_impl_config.go:60`): a bad scratch-space just logs a Warn and reverts to heap — reintroducing the OOM this PR exists to prevent. Consider hard-failing when mmap was explicitly requested.
- **`Get` nil-panic message is broken** (`:171`): `panic(fmt.Sprintf(... b.storeName))` dereferences `b` in the `b == nil` branch → a generic nil panic instead of the intended message. Cosmetic.
- **`sweepOrphanMmapFiles`** only runs on tier2 (`WithStoresScratchSpace`), not tier1, and is a no-op guard when scratch == "" (globs cwd). Tier1 SIGKILL orphans won't be swept. Low — files are unlinked at open, so orphans are rare.
- **Test gaps**: no concurrency/race test (#3), no memory-backend write-during-snapshot test (#4), no test that a `FinalStoreMap` mid-load error closes already-loaded stores (it *is* handled — detach is skipped, so deferred `Stages.Close()` catches them — but untested).

### Verified clean (candidates that turned out fine)
- `merge.go:88` partial-store iterate → **copies** the value. ✅
- Memory `Load` (`:115`) clones the key; the value is already freshly copied by `UnmarshalIter` (`vtproto.go:245`). ✅
- `tools/` `Iter` callers use the value in-callback only. ✅
- Store close lifecycle: ownership detach in `FinalStoreMap`, `defer stores.Close()` in tier1, `fileWriter.Write` defers `reader.Close()` → `snap.Close()`. No double-close/leak on success or `FinalStoreMap` error paths. ✅
- Snapshot write-gate + Close-idempotency + use-after-close **are** tested (`kv_impl_snapshot_test.go`). ✅

**Top action: resolve #1 (default backend) before merge** — it silently changes behavior for every deployment.

sduchesneau (approved):
LGTM

PR #851, opened 2026-07-20
Improve eth_call / WASM extension call metrics

maoueh:
## Why

Poor RPC performance — mainly `eth_call`, a WASM extension call — can clog tier1/tier2 request processing, and today there is almost no way to spot it. The only signals are the per-request stats log (which merged every extension into a single duration) and the client-facing `ModuleStats`. Neither can be dashboarded or alerted on.

This adds operator-facing metrics and enriches the request stats log, and fixes two latent bugs found while mapping the code.

## What

**Bug fixes (own commits, independently validatable)**

- **Error-path leak** — a WASM extension call that returned an error panicked before `RecordModuleWasmExternalCallEnd` ran, so the in-process entry was never removed. `updateDurations` keeps adding `time.Since(start)` for every in-process entry, so a single failed `eth_call` kept inflating the reported external call duration for the rest of the request. Fixed in both the `wazero` and `wasmtime` runtimes.
- **Shared-cache gatherer** — `WasmMetricsGatherer` built a new metric when none existed for a `(module, extension)` pair but never stored it back (and never initialized its maps), so external call metrics on the shared-cache path were silently dropped. Also nil-guarded the logger, which the shared cache leaves unset.

**Part A — Prometheus metrics**

Per tier, following the existing name-prefix convention (no `tier` label):

- `substreams_tier{1,2}_wasm_extension_call_counter{extension,outcome}`
- `substreams_tier{1,2}_wasm_extension_call_duration_seconds{extension,outcome}`

`extension` e.g. `eth:call`, `outcome` = `success`|`error`. Duration buckets extend the Prometheus defaults with a 30s/60s tail so timeouts stay distinguishable. Recorded at the single chokepoint every extension call flows through, in both runtimes.

**Part B — request stats log breakdown**

- `wasm_ext_call_metrics` — per extension: `count`, `total_ms`, `avg_ms`, `max_ms`.
- `wasm_ext_call_metrics_by_module` — same, per module + extension (only modules that actually called out; empty otherwise).

`max_ms` surfaces the single slow call an average hides. The pre-existing `module_wasm_ext_duration` is unchanged; the new fields are additive and placed after the fields consumed by reporting systems.

## Naming

The recorded event is a generic WASM extension call — `eth_call` is only the common case — so metrics/helpers are named `wasm_extension_call` rather than `rpc`, keeping the chain-agnostic chokepoint chain-agnostic.

## Testing

- Unit tests for the tier routing helper (routing, outcome separation, nil-guard), the per-extension/per-module aggregation (across local + running + completed sources, empty case, divide-by-zero guard, `max_ms` tracking), and the shared-cache gatherer fix (verified to fail without the fix).
- Full suite green.

Design doc: `docs/superpowers/specs/2026-07-20-eth-call-rpc-metrics-design.md` (locally gitignored, not in this branch).

Draft while I test on staging.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:89ea94c</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:89ea94c</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:2a99276d472357056e693c61d819b3eb85d932dcd03fc0153c24fb9a0d9f211c</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

billettc (approved):

PR #864, opened 2026-08-05
fix(tier1): restart when the hub can no longer link live blocks

YaroShkvorets:
Pass `hub.WithMaxConsecutiveUnlinkableBlocks(5)` option so the ForkableHub to make it restart after 5 unlinkable blocks. Otherwise it could get stuck forever requiring manual restart.

Similar to this relayer fix: streamingfast/firehose-core@4696483

### DIFF
--- a/app/tier1.go
+++ b/app/tier1.go
@@ -33,6 +33,10 @@ import (
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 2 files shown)


sduchesneau (approved):
LGTM same as relayer

PR #866, opened 2026-08-09
Log the worker counts of an incoming tier1 request

maoueh:
## Why

A client asking for 300 parallel workers and getting 15 leaves no trace of the negotiation anywhere in the logs today, so the gap can't be diagnosed after the fact. Three separate things can cap the worker count and none of them are observable:

1. **Plan/auth negotiation** — the trusted header from the authentication layer sets the ceiling, the client header can only lower it. Both inputs collapse into a single number that is never logged.
2. **Worker pool contention** — `SessionWorkerPool.Borrow` can come back empty when the pool is shared with other requests. Only visible at debug level.
3. **Not enough work** — a request with fewer segments than workers can never use them all.

## What changed

**`incoming Substreams Blocks request`** gained a `parallelism` object:

```
parallelism: {requested_workers: 300, granted_workers: 15, workers: 15,
              workers_source: "trusted_header", plan_tier: "SCALING",
              stage_layer_executors: 5}
parallel_segment_count: 4213
stage_count: 3
```

`workers_source` tells which input won, so "granted 15 because of the plan tier" and "lowered to 15 because the client asked for it" are no longer the same log line.

**`substreams request stats`** gained a `workers` object, tier1 only:

```
workers: {requested: 300, granted: 15, effective: 15, peak: 4,
          pool_exhausted_count: 87, pool_rampup_deferred_count: 2}
```

`peak` well below `effective` together with a high `pool_exhausted_count` points at pool contention rather than at the negotiated limit.

**`GetEffectiveHeaderValues`** now returns an `EffectiveParallelism` struct instead of two bare integers — that is what makes the negotiation inputs available for logging. The arithmetic is unchanged, and the existing behavior is still covered by the table test plus new cases for the client-asks-for-more and legacy-header paths.

## Notes

- No proto change, nothing new sent to clients — `SessionInit.MaxParallelWorkers` already carries the effective count.
- The six worker counters are atomic rather than guarded by the `Stats` mutex, so the scheduler loop doesn't contend with the block-processing path for a counter bump.
- The scheduler reads the stats through a new nil-safe `reqctx.ReqStatsOrNil`, since `reqctx.ReqStats` panics where no stats object is installed.

## Testing

`go test ./...` passes, and `go test -race ./metrics/...` is clean.

### DIFF
--- a/docs/release-notes/change-log.md
+++ b/docs/release-notes/change-log.md
@@ -11,6 +11,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Added
+
+- Server: tier1 request logs now explain how many parallel workers a request actually got, and why. A client asking for 300 workers and getting 15 previously left no trace of the negotiation anywhere in the logs.
+
+  The `incoming Substreams Blocks request` entry gained a `parallelism` object (`requested_workers` as asked by the client, `granted_workers` as allowed by the authentication layer, the effective `workers`, `workers_source` telling which of the two applied, plus `plan_tier` and `stage_layer_executors`), along with `parallel_segment_count` and `stage_count` — a request with fewer segments than workers can never use them all.
+
+  The `substreams request stats` entry gained a `workers` object (`requested`, `granted`, `effective`, `peak`, `pool_exhausted_count`, `pool_rampup_deferred_count`), reported on tier1 only. A high `pool_exhausted_count` with a `peak` well below `effective` means the shared worker pool ran dry, a case that was previously only visible at debug level.
+
 ### Fixed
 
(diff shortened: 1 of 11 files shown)


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:a60c14b</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:a60c14b</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:ea78e07eec2f68e1d21f78b7f8a5b3c99929159ecd79dccf589e923077b0adae</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

sduchesneau (changes requested):
add it to the new `substreams request progress` message too

sduchesneau (approved):
LGTM

PR #867, opened 2026-08-10
fix(sink-sql): three correctness fixes in the from-proto sink

sduchesneau:
Three independent correctness fixes in the from-proto SQL sink. Each is one commit and stands alone; they are together only because they touch adjacent code.

### 1. `bytes` fields were corrupted in `BYTEA` columns

Under `--bytes-encoding=raw`, the default, both inserters mangled them, each differently:

```
input   de ad be ef 00 7f ff
before  accumulator  "deadbeef007fff"   hex text, 2x the bytes
        row inserter "'3q2+7wB//w=='"   base64, quotes included
after   both         de ad be ef 00 7f ff
```

`ValueToString` rendered `E'<hex>'::BYTEA`, which casts the hex *text* to bytea rather than decoding it. `RowInserter` binds parameters instead of interpolating, so rendering any literal there stores that literal's characters. The same confusion broke array parameters, where each element kept its SQL quotes (`'alpha'` rather than `alpha`).

Nothing failed loudly: the rows were all there, and every query against those columns matched nothing.

### 2. An `enum` field crashed the sink

protoreflect yields a `protoreflect.EnumNumber`, a named `int32` that neither dialect's type switch matched — PostgreSQL panicked with `unsupported type: protoreflect.EnumNumber`, ClickHouse on a failed `value.(int32)`. The two disagree on the column type (`TEXT` versus `Int32`) and the walk is shared, so it now carries both renderings and each dialect takes the one matching the column it created.

Found by pointing the sink at `erc20-balance-changes` from substreams.dev, which reproduces it on the first block.

### 3. A bounded run dropped the blocks it was still holding

Blocks accumulate until `--block-batch-size` of them have gone by, and nothing drained that buffer at the end of the range. A run ending mid-batch silently dropped its last blocks along with the cursor covering them, and still reported success. With a batch size larger than the requested range, nothing was written at all.

The base sinker already offers `SinkerCompletionHandler` for this and `db_changes` implements it; this does the same for `db_proto`.

### Impact

Databases populated by an affected version hold corrupted values in `bytes` and repeated-scalar columns, and may be missing the tail of a bounded run. The affected ranges need re-syncing.

### Tests

New integration tests cover each: a bytes/timestamp/array round trip through both inserters, an enum walked through `dynamicpb`, and a bounded run whose batch size exceeds its range. Each was confirmed to fail against the previous code.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:788cdea</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:788cdea</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:8fc225de26ab19a9bbaf87f6c54e0f968adf18d7d11da3cbcff0c0e608a39028</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

sduchesneau (commented):
double checked everything, seems good to me, if someone else can vet!

maoueh (approved):

PR #868, opened 2026-08-10
Decode blocks on a worker pool in the from-proto sinker

sduchesneau:
### The sink, not PostgreSQL, was the bottleneck

Unmarshalling into `dynamicpb` and walking the message descriptor is where the from-proto sink spends its time, and it ran on the goroutine draining the stream. Measured with no database involved at all:

| stage | narrow entity (2 cols) | wide entity (60+ cols) |
|---|--:|--:|
| unmarshal only | 1.79M rows/s | 203k rows/s |
| + walk | 855k rows/s | 128k rows/s |
| + encode | 745k rows/s | **96k rows/s** |

Binary COPY absorbs ~800k rows/s, so for wide entities PostgreSQL had roughly eight times the headroom the single-threaded handler could reach.

### What this does

Blocks are independent, so decode now runs on a worker pool. Each worker fills a `BufferedInserter`; the flush goroutine replays them **in block order**, inside the same transaction as before, so the database sees exactly the sequence a serial walk produced. Decoding happens before the transaction opens, since it touches no database state and would otherwise pin a connection for the CPU-bound work.

Scaling on a 16-core machine:

| workers | rows/s | speedup |
|--:|--:|--:|
| 1 | 129k | 1.00x |
| 8 | 534k | **4.13x** |
| 15 | 548k | 4.24x |

Flat past eight — the work is allocator-bound before it runs out of cores — so the default is one per core less one, **capped at eight**. `DecodeWorkers` overrides it.

### It replaces an unsound path rather than adding to one

`SinkerFactoryOptions.Parallel` was hardcoded `false` at every call site, so the existing parallel path was unreachable. Had it run it was unsound: `Database.Clone()` overwrote the receiver's embedded struct and returned the receiver, so every goroutine shared one `*sql.Tx` and raced on it through `BeginTransaction`/`CommitTransaction`, the clone was not used for the inserts anyway, and errors were appended to a shared `appengine.MultiError` without a lock. `Clone()` goes with it, along with the `google.golang.org/appengine` import that only existed to carry that error slice, and `holding` moves from a package-level global into `Sinker`.

Also puts the zap debug fields on the walk behind `Check`. zap builds its `Field` arguments whether or not the level is enabled, so `zap.Any` on a table info struct and on every row's value slice allocated on every call — worth ~6% on a narrow entity.

### Tests

`TestDbProtoPostgresDecodeWorkers` runs the same eight-block stream through one worker and through eight and compares the tables in physical row order, plus an absolute check that blocks land ascending — a differential test alone cannot see an ordering mistake both runs share. Verified to fail when the replay order is reversed.

### DIFF
--- a/cmd/substreams/sink_sql_common.go
+++ b/cmd/substreams/sink_sql_common.go
@@ -341,7 +341,6 @@ func runFromProtoSink(cmd *cobra.Command, driver, manifestPath, dsnString string
 		UseConstraints:  useConstraints,
 		UseTransactions: true,
 		BlockBatchSize:  blockBatchSize,
-		Parallel:        false,
 		Encoding:        encoding,
 		Clickhouse:      fromProtoClickhouseOptions(cmd, driver),
 	})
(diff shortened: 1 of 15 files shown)


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:6070334</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:6070334</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:a5a8ca2349d1d87674df877a50f2ebe9a21ad745ddf6ed7691620c7f6372925d</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

sduchesneau:
@copilot resolve conflicts

sduchesneau (commented):
Reviewed, Claude work, looks sane

maoueh (approved):

PR #869, opened 2026-08-10
Sink from-proto through a disk spool, and load without constraints

sduchesneau:
The from-proto SQL sink wrote to the database on the stream's own goroutine, with the
constraints in place while it loaded, through a flag surface that had grown to say
different things depending on the driver and on where the stream was. This changes all
three, and adds the index the reorg path always needed.

## What changes

**Rows go through a spool.** They land on local disk, pre-encoded into whatever the target
can take unchanged, and a background goroutine applies whole segments. The stream stops
waiting on the database, and blocks already downloaded survive a restart instead of being
streamed — and paid for — twice. Against `erc20-balance-changes` over 50,000 Ethereum
mainnet blocks (17M rows) into a containerised PostgreSQL 17, this took the run from 190.9s
to 70.9s and the stream from 262 to 673 messages per second, with an identical table.

Both drivers spool. ClickHouse keeps the guarantees it already had — no transactions, its
cursor in a file — so applying a segment there is the same inserts followed by the same
cursor write, in the same order.

**Constraints are created after the load, not during it.** Measured through binary COPY
over 500k rows, loading with foreign keys in place costs 27.7x against 3.3x for building
the same ones afterwards. `--apply-constraints` says when: `auto` (default) has the sink do
it when the stream reaches chain HEAD, `manual` leaves it to `sink postgres constraints
apply`, `always` keeps the old behaviour. `auto` is the default despite being a
stop-the-world pass — a backfill that ends with no primary keys and no foreign keys is a
silent wrong result that looks like success.

A stop block deliberately does *not* trigger it. Reaching `-t` ends the run without saying
the backfill is done — a range is routinely one chunk of several — and building the
constraints there would leave every later chunk loading into a constrained schema, the
27.7x case this exists to avoid. The run says on the way out what it left behind.

Each statement commits on its own, so a run killed by the OOM killer keeps what it
finished. `--constraints-parallelism` runs them side by side (keys first, then the foreign
keys that reference them), and `--constraints-work-mem` sets `maintenance_work_mem` for the
duration of each: at the usual 64MB default an index build over a large table spills to an
external merge sort.

**`_block_number_` is indexed, at startup, concurrently.** Every table carries that column
and every reorg deletes from every table by it, and a foreign key indexes only its
referenced side — so each undo was a sequential scan per table. Measured over 10GiB: 2.6s
to build, 218MiB against 10GiB, and one table's undo goes from 1.296s to 15ms. It is
deliberately *not* part of the constraint pass: those describe the schema and are the
operator's to schedule, where this one the sink depends on.

**`--write-mode` replaces guessing.** Which write path ran used to be derived from whether
a directory flag was set and whether the schema's foreign keys happened to form a cycle,
and an unorderable schema was silently downgraded to row-at-a-time inserts — an order of
magnitude slower, announced only in a log line. It is now `copy`, `batch-insert`,
`row-insert` or `auto`, and an explicit mode the driver or schema cannot support is an
error that names the fix.

**Flags are named after what they spend.** `--decode-*` is CPU, `--db-write-*` is one
commit to the database, `--spool-*` is how far ahead of it the stream may run. Commit size
is not a setting: the sink measures each one and sizes the next segment toward
`--db-write-target-duration`, block payloads differing by orders of magnitude across chains
being exactly what makes a fixed block count give unstable durations.

**`substreams tools extract-proto --sql`** writes a module's output proto back out with
every SQL annotation commented out beside the message and field it applies to, and the
annotations file next to it so the result parses as-is. Starting from a package whose
author never annotated it was otherwise a matter of finding the right proto and the right
extension names by hand.

**ClickHouse takes a package with no annotations.** It required `order_by_fields` on every
table, so a Substreams not written for the sink could neither be set up nor run. Tables now
default to `ORDER BY (_block_number_, _row_id_)`, `PRIMARY KEY (_block_number_)` and
`PARTITION BY (toYYYYMM(_block_timestamp_))`. `_row_id_` numbers the rows one block writes
to one table: the tables are `ReplacingMergeTree`, so a sorting key that cannot tell those
rows apart collapses a whole block into one row at merge time. It is added only where no
`order_by_fields` is declared, and PostgreSQL is untouched. Three things were in the way
behind the error message — the ClickHouse database was built with "use proto options"
hardcoded to true, so an unannotated package inserted nothing at all; `setup` exited with
`Flag "sink-info-folder" does not exist`, those flags being registered on the run command
only; and an undo wrote tombstones without `_row_id_`, landing them on a different sorting
key where they removed nothing. `erc20-balance-changes` / `map_balance_changes` now fills a
ClickHouse database as-is.

## Reviewing

Fourteen commits, each building and vetting clean on its own. The first four are the shape
of it:

| commit | what |
|---|---|
| `Encode rows in PostgreSQL's binary COPY wire format` | leaf package, no callers yet |
| `Hold rows on disk and apply whole segments in the background` | the spool: segment lifecycle, disk budget, sizing. `Codec` and `Applier` are the driver's business |
| `Write through the spool, load without constraints` | the behaviour change and the flag surface — the one to read closely |
| `Cover the spool, the write modes and the constraint timings` | container-backed tests |

The rest are fixes and follow-ups found while using it, each self-contained: draining the
spool before the stop block, the startup report of missing constraints, committing the
constraint pass in pieces, the module argument on the schema commands, and the block number
index with its measurement.

## Compatibility

`--block-batch-size` and `--no-constraints` still work for a release and warn. Everything
else here is new, so nothing else needs an alias. A from-proto flag typed for a
DatabaseChanges Substreams is now an error rather than silently ignored, and so is the
reverse — a DatabaseChanges Substreams owns its schema, so `sink postgres constraints`
refuses it outright.

`setup` and `constraints` take the output module as an optional second argument, as the run
command always has; a package with more than one candidate had no way to say which.

### Commands

- [ ] add [<module>] optional argument to sink postgres|clickhouse setup <manifest> — was inferred from the package with no way to override
- [ ] add sink postgres constraints apply <manifest> [<module>], replacing sink postgres apply-constraints <manifest>
- [ ] add sink postgres constraints drop <manifest> [<module>] — escape hatch after --apply-constraints=always, or before resuming a backfill
- [ ] add substreams tools extract-proto [<manifest> [<module>]] — writes a module's output proto back out, with --sql and --output-dir
- [ ] --sql annotates every message/field with commented-out (schema.table)/(schema.field) options and writes schema.proto beside it so it parses as-is

### Flags added

- [ ] add --write-mode=auto|copy|batch-insert|row-insert — how a sealed spool segment reaches the database; unsupported explicit mode is an error, not a downgrade
- [ ] add --decode-workers (0 = one per core less one, capped at 8) — CPU only, does not change what the database sees
- [ ] add --decode-batch-size (0 = 4 × workers) — blocks held in memory and decoded together
- [ ] add --db-write-target-duration (3s) — how long one commit should take; the sizer measures each and sizes the next segment toward it
- [ ] add --db-write-max-size (512MiB) — ceiling on the segment size the sizer may choose
- [ ] add --spool-dir (./localdata/spool) — where pending segments are written
- [ ] add --spool-max-size (8GiB) — disk budget, and the only bound on how far ahead of the database the stream may run
- [ ] add --spool-max-idle (10s) — commit the open segment when no new row has reached it for that long; 0 disables
- [ ] add --apply-constraints=auto|manual|always (default auto) — when the schema's constraints are created
- [ ] add --disable-foreign-keys, --disable-primary-keys=<tables|all>, --disable-unique-constraints=<tables|all> — which constraints the schema declares
- [ ] add --disable-block-number-index — leave out the _block_number_ index; only sensible for a run that can never reorg
- [ ] add --constraints-per-transaction (1), on constraints apply|drop only — how many are committed at a time

### Flags deprecated / moved

- [ ] deprecate --block-batch-size in favour of --decode-batch-size; honoured one release with a warning
- [ ] deprecate --no-constraints in favour of --disable-foreign-keys --disable-primary-keys=all --disable-unique-constraints=all (it had been deleted outright on this branch)
- [ ] stop registering the from-proto run flags on setup and on the constraints commands — they never decode a block or write a row
- [ ] group the run command's --help under From-proto mode flags: / DatabaseChanges mode flags: / Flags (both modes):, and drop the now-redundant [mode] prefixes

### Mode guards

- [ ] error on any from-proto flag typed for a DatabaseChanges Substreams (Flags().Changed, so defaults never trip it)
- [ ] error on any DatabaseChanges flag typed for a from-proto Substreams — was a warning on setup only
- [ ] sink postgres constraints apply|drop refuses a DatabaseChanges module outright: its schema comes from the manifest's schema.sql

### Spool

- [ ] add sink/sql/db_proto/sql/spool — owns the segment lifecycle, disk budget and sizing; Codec says what the bytes are, Applier how they land
- [ ] add four on-disk formats: binary COPY per table, rendered SQL tuples per table, one interleaved row log in walk order, typed values for ClickHouse
- [ ] row-insert gets the interleaved log because a cyclic FK graph has no table order that keeps a parent ahead of its children — only the walk does
- [ ] give ClickHouse the spool, on the guarantees it already had: same inserts, same cursor-file write, same order; no dedup token, no cursor table, no migration
- [ ] remove the hardcoded QueueDepth: 2, which capped the spool at ~2GiB and made the 8GiB budget unreachable; the byte quota is now the only ceiling
- [ ] check the disk quota before the manifest is written rather than after, and count the open segment against it
- [ ] seal the open segment on idle, from a timer goroutine, with a mutex on current — a stalled sinker is blocked in a read and cannot check anything
- [ ] persist a cursor-only segment when a flush carries no rows, so a long empty stretch is not re-streamed on restart
- [ ] generalise Buffer.resize into a shared sizer: one dial (segment bytes) in every write mode, floor an internal constant rather than a flag

### Constraints

- [ ] load without constraints and create them afterwards — 27.7x vs 3.3x through binary COPY over 500k rows
- [ ] order table operations by a topological sort over the foreign keys, so the fast paths work with constraints in place
- [ ] undo a reorg by deleting from every table explicitly, children first, instead of relying on a cascade that only exists when constraints do
- [ ] commit the constraint pass one statement at a time instead of one transaction for all of it — an OOM used to lose the entire pass
- [ ] key constraints by (relation, name), not name: fk_block exists on every table, so one present anywhere counted as present everywhere
- [ ] fix constraints drop failing with cannot drop constraint block_pk ... other objects depend on it — same root cause
- [ ] warn at startup naming the constraints the schema is meant to have and does not, with what will create them
- [ ] warn instead of erroring when the output has no schema.proto annotations, on both the run and constraints apply

### Block number index

- [ ] create an index on _block_number_ on every table — a foreign key indexes its referenced side only, so every undo was a sequential scan per table
- [ ] create it at sink startup and CONCURRENTLY, decoupled from the constraint pass: that runs in transactions, which a concurrent build cannot
- [ ] drop and rebuild an index a previous interrupted concurrent build left invalid — IF NOT EXISTS would otherwise keep it forever

### Behaviour fixes

- [ ] drain the spool at a bounded run's stop block — it panicked on a nil transaction — and leave the constraints alone there, a stop block ending the run without saying the backfill is done
- [ ] switch to direct inserts at chain head: drain the spool, close it for good, and say which flags stop apply
- [ ] flush the blocks still held when a bounded run reaches its stop block — they were dropped along with their cursor, and the run reported success

### Observability

- [ ] report downloaded_through / applied_through / blocks_ahead / blocks_buffered / peak_blocks_ahead; past half the spool's disk budget it logs at warning level, a block count saying nothing once rows are on disk
- [ ] log the resolved write mode, decode workers, batch size, spool settings and constraint policy in one line
- [ ] log why auto resolved to row-insert, and warn when --apply-constraints=always picks the measured 27x path

### Tests and measurements

- [ ] add TestDbProtoPostgresWriteModes — the three PostgreSQL modes must produce identical tables
- [ ] add TestDbProtoClickhouseSpool — a spooled ClickHouse run must match an unspooled one
- [ ] add TestDbProtoConstraintsParallelism, TestDbProtoPostgresMissingConstraints, TestDbProtoPostgresBlockNumberIndex, TestDbProtoConstraintsAtEndOfRangeWithSpool
- [ ] add TestConstraintTarget / TestConstraintTargetDistinguishesTables — same name on two tables must be two c
- [ ] add TestRenderProtoFileParsesWithAnnotations — the scaffold must compile once its options are uncommented
- [ ] add the benchmarks package: TestBlockNumberIndexCost, TestConstraintCost, TestCopyVsInsert, client-ceilingus live-benchmark.sh and its README
- [ ] measured: index costs 2.6s on a 44s load and 218MiB against 10GiB, and takes one table's undo from 1.296s to 15ms

### Docs

- [ ] rewrite docs/references/sql/proto-annotations.md: flag groups, write modes, constraint timings, the block index, and starting from an unannotated package
- [ ] changelog entries for the spool, the write modes, the flag rename, the mode guards, the constraint pass, t

### ClickHouse without proto annotations

- [ ] default a table with no 'order_by_fields' to ORDER BY (_block_number_, _row_id_), PRIMARY KEY (_block_number_), PARTITION BY (toYYYYMM(_block_timestamp_)) — a declared primary key leads the key, ClickHouse requiring it as a prefix
- [ ] add the _row_id_ column: the count of rows the block has already written to that table, threaded through the walk in a per-block map rather than held on the database, the decoder walking blocks on parallel goroutines
- [ ] pass options.UseProtoOption to the ClickHouse database instead of a hardcoded true — an unannotated package was walked and inserted nothing
- [ ] register the ClickHouse state flags on setup as well, which exited with Flag "sink-info-folder" does not exist
- [ ] carry _row_id_ into the undo tombstone, without which it lands on a different sorting key and removes nothing
- [ ] refuse to start on a database whose tables disagree with the package about _row_id_ — annotating a message after the fact changes the sorting key of a table that already holds rows, and CREATE TABLE IF NOT EXISTS would keep the old one and write values into the wrong columns
- [ ] add TestDbProtoClickhouseWithoutAnnotations and ...Undo against a real server: three rows survive OPTIMIZE FINAL with distinct ids, an undo removes exactly the undone block's rows, and a mismatched table is rejected
- [ ] add unit cover for the DDL defaults and for the numbering, including that it resets per block and is absent on PostgreSQL

---

## Measured end to end

Uniswap v4 on Base, `map_events`, blocks 25,350,988 → 36,712,209 (11,361,221 blocks), into
containerised ClickHouse and PostgreSQL 17. Both engines wrote **370,507,376 rows** and
agreed on every table to the row, which is the correctness result as much as the timing one.

| | PostgreSQL (no constraints) | PostgreSQL (w/constraints) | ClickHouse |
|---|---:|---:|---:|
| wall time | 18m 36s | 37m 08s | 23m 04s |
| — fill | 18m 36s | 18m 36s | 23m 04s |
| — constraint pass | — | 18m 32s | n/a |
| rows/s (fill) | ~332k | ~332k | ~268k |
| database size | 107 GB | 138 GB | — |

The constraint pass costing as much as loading all 370M rows is what makes the stop-block
change worth having: a ten-chunk backfill used to pay that at every chunk boundary, and
have every chunk after the first load the 27.7x path.

Per-block sink cost is now small enough that the run is usually delivery-bound: block
insert 0.02ms, entities 0.01ms, flush 0.00ms, with 77% of wall time waiting on the stream
over the full range.

## Fixes folded in

#881 merged into this branch: twelve bugs found reviewing the work above, each with a
regression test checked to fail without its fix — among them the ClickHouse cursor never
reaching its file while spooling, `_blocks_` rows racing the applier, a descriptor leaked
per sealed COPY segment, reorgs failing on cyclic schemas, backslashes doubled in rendered
literals, and empty or enum arrays aborting a segment in the default write mode.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:b5e97ac</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:b5e97ac</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:93f147d56dc616e61f337d89cc77fb85180fd9259c54698a30436efe055ef17d</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

sink/sql/db_proto/benchmarks/correctness_test.go
```diff
@@ -0,0 +1,148 @@
...
+	tcpostgres "github.com/testcontainers/testcontainers-go/modules/postgres"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+// TestPgCopyBinaryRoundTrip is the safety net for the binary encoder: a value written
+// through pgcopy must come back out of the server byte-identical to what a
+// parameterised INSERT of the same value produces. Binary COPY does no coercion, so an
+// encoder bug shows up as either a hard COPY failure or, worse, silently wrong data.
+func TestPgCopyBinaryRoundTrip(t *testing.T) {
+	requireDocker(t)
```
maoueh: Tests that are relatively fast to perform, even within Docker, should all be run by default. If this can easily launch testcontainers to make the test, it should be enabled automatically.

sink/sql/db_proto/sql/postgres/buffer/segment.go
```diff
@@ -0,0 +1,241 @@
...
+
+func sortedKeys[V any](in map[string]V) []string {
+	out := make([]string, 0, len(in))
+	for key := range in {
+		out = append(out, key)
+	}
+	sort.Strings(out)
+
+	return out
+}
```
maoueh: Doesnt maps + slices could do that in one line?

sink/sql/db_proto/sql/postgres/buffer_inserter.go
```diff
@@ -0,0 +1,95 @@
...
+	"go.uber.org/zap"
+)
+
+// bufferInserter routes rows into the on-disk buffer instead of the database.
+//
+// It sits behind the same pgInserter/pgFlusher interfaces as the accumulator, so the
+// sinker is unchanged: what differs is that a "flush" only seals a segment, and the
+// database is written to later, by the buffer's own goroutine. That is the whole point —
+// the stream stops waiting on PostgreSQL.
+type bufferInserter struct {
```
maoueh: We have also now BufferedInserter, not sure if we could not find a better name

sink/sql/db_proto/stats/progress.go
```diff
@@ -0,0 +1,139 @@
...
+	value := float64(n)
+	for _, unit := range []string{"B", "KiB", "MiB", "GiB"} {
+		if value < 1024 {
+			return fmt.Sprintf("%.1f%s", value, unit)
+		}
+		value /= 1024
+	}
+
+	return fmt.Sprintf("%.1fTiB", value)
+}
```
maoueh, line 163: I'm adding a formatx package to hold all this, once existing, will point Agents.md to it.

maoueh (dismissed):

cmd/substreams/sink_postgres_generate_csv.go
```diff
@@ -104,7 +104,7 @@ func sinkPostgresGenerateCSVE(cmd *cobra.Command, args []string) error {
 		supportedOutputTypes,
 		manifestPath,
 		outputModule,
-		"sink_database_changes",
```
maoueh, line 107: Yeah!

bin/test.sh
```diff
@@ -18,10 +18,15 @@ main() {
...
+
+  # tests_e2e is its own Go module, so `go test ./...` above never reaches it. It stays
+  # out of this script until it passes on Linux: the node writes root-owned files into
+  # the bind-mounted t.TempDir, which fails every container test's cleanup on a CI
+  # runner. See the fix/tests-e2e-on-linux branch.
+  #
   # pushd tests_e2e &> /dev/null
   # go test ./... "$@"
-  # popd
+  # popd &> /dev/null
```
maoueh, line 29: Useless, still not running ...

cmd/substreams/sink_postgres.go
```diff
@@ -29,27 +29,96 @@ var sinkPostgresSetupCmd = &cobra.Command{
...
+	setModeGroupedUsage(sinkPostgresCmd)
 
 	setupFlags := sinkPostgresSetupCmd.Flags()
 	addCursorTableFlags(setupFlags)
 	addOnModuleHashMismatchFlag(setupFlags)
-	setupFlags.Bool("postgraphile", false, "[DatabaseChanges mode] Will append the necessary 'comments' on cursors table to fully support postgraphile")
-	setupFlags.Bool("system-tables-only", false, "[DatabaseChanges mode] will only create/update the systems tables (cursors, substreams_history) and ignore the schema from the manifest")
-	setupFlags.Bool("ignore-duplicate-table-errors", false, "[DatabaseChanges mode][Dev] Use this if you want to ignore duplicate table errors, take caution that this means the 'schema.sql' file will not have run fully!")
+	setupFlags.Bool("postgraphile", false, "Will append the necessary 'comments' on cursors table to fully support postgraphile")
+	setupFlags.Bool("system-tables-only", false, "will only create/update the systems tables (cursors, substreams_history) and ignore the schema from the manifest")
```
maoueh, line 101: Why we lost `[DatabaseChanges]` note, if it's because it's not adding system tables in both, we should list all possible system tables for both mode.

cmd/substreams/sink_sql_common.go
```diff
@@ -125,6 +128,63 @@ func boolFlag(cmd *cobra.Command, name string) bool {
...
 
+// stringSliceFlag mirrors boolFlag for the per-table constraint switches, which are only
+// registered for the from-proto commands.
+func stringFlag(cmd *cobra.Command, name string) string {
+	if sflags.FlagDefined(cmd, name) {
+		return sflags.MustGetString(cmd, name)
+	}
+
+	return ""
+}
```
maoueh, line 139: I'm thinking of adding those generated version on sflags now, maybe `OptionalGetString` with default, thinking about it.

cmd/substreams/sink_postgres_generate_csv.go
```diff
@@ -104,7 +104,7 @@ func sinkPostgresGenerateCSVE(cmd *cobra.Command, args []string) error {
...
+		sinkUserAgent("sink_database_changes", sinkPostgresDriver),
```
maoueh, line 107: `sink_database_changes` and even `sinkUserAgent` don't really use more "standard" agent formats which uses a lot the `/` for namespacing.

We should do the same + show full name `postgres`, `clickhouse`.

maoueh: We should also embedded CLI version in the user agent too

cmd/substreams/sink_sql_common.go
```diff
@@ -138,43 +198,174 @@ func isDatabaseChangesType(outputType string) bool {
 // addDatabaseChangesModeRunFlags registers the run flags that only apply when the
 // selected module outputs DatabaseChanges.
 func addDatabaseChangesModeRunFlags(flags *pflag.FlagSet) {
-	flags.Int("undo-buffer-size", 0, "[DatabaseChanges mode] If non-zero, handling of reorgs in the database is disabled. Instead, a buffer is introduced to only process blocks once they have been confirmed by that many blocks, introducing a latency but slightly reducing the load on the database when close to head. Set to 0 to enable reorg handling in the database (required for some databases like Postgres).")
```
maoueh, line 141: Are we sure some of them `[DatabaseChanges mode]` should keep those?

cmd/substreams/sink_sql_common.go
```diff
@@ -138,43 +198,174 @@ func isDatabaseChangesType(outputType string) bool {
...
+	flags.Int("constraints-parallelism", 1, "How many constraints are created or dropped at once. They go on independent relations, so the server can build them side by side, and the only ordering that matters is that a foreign key needs the key it references to exist — which the pass handles by running the keys first. One at a time by default, since each build takes its own --constraints-work-mem and holds a lock on its table. On a schema whose rows sit in one dominant table this buys little; on one with several large tables it is close to linear.")
+	flags.String("constraints-work-mem", "", "What maintenance_work_mem is set to for the duration of each constraint statement, e.g. 1GB. Empty leaves the server's own setting alone, which on most is 64MB — at which an index build over a large table spills to an external merge sort. Raising it for the pass alone is the cheapest thing that makes it faster. It multiplies with --constraints-parallelism, each concurrent build taking its own.")
+
+	flags.Int("constraints-per-transaction", 1, "Deprecated, use --constraints-parallelism.")
+	_ = flags.MarkDeprecated("constraints-per-transaction", "use --constraints-parallelism")
+}
```
maoueh: Those never existed no? So just remove?

cmd/substreams/sink_sql_common.go
```diff
@@ -138,43 +198,174 @@ func isDatabaseChangesType(outputType string) bool {
...
+	flags.StringSlice("disable-primary-keys", nil, "Tables that go without a primary key, or 'all'. No primary key also means no index on the entity id.")
+	flags.StringSlice("disable-unique-constraints", nil, "Tables whose unique constraints are left out, or 'all'.")
+	flags.Bool("disable-block-number-index", false, "Leave out the index on _block_number_. Every table carries that column and every reorg deletes from every table by it, so without the index each undo is a sequential scan per table — a foreign key indexes its referenced side only. It is created when the sink starts, concurrently and outside the constraint pass: --apply-constraints describes the schema and is yours to schedule, where this one the sink depends on to undo a reorg. Measured over 10GiB it costs 2.6s to build and 2% of the table. Only dead weight on a run that can never reorg, such as --final-blocks-only.")
+	flags.String("proto-file-override", "", "Override protobuf file to use instead of extracting from substreams package")
+
+	flags.Bool("no-constraints", false, "Deprecated, use --disable-foreign-keys --disable-primary-keys=all --disable-unique-constraints=all.")
```
maoueh: Don't we want to notify the new `--apply-constraints=...` instead?

cmd/substreams/sink_sql_common.go
```diff
@@ -320,30 +526,55 @@ func runFromProtoSink(cmd *cobra.Command, driver, manifestPath, dsnString string
...
 	}
 	if !useProtoOption {
-		useConstraints = false
+		// Without the schema annotations there are no relations to constrain. The index on
+		// _block_number_ is not declared by them either, so it stays.
+		constraints = protosql.DisableAllConstraints().WithBlockNumberIndex(constraints)
+
+		zlog.Warn("the module's output carries no schema.proto annotations, so its tables get no primary keys, no unique constraints and no foreign keys. " +
+			"Queries against them are sequential scans and duplicate ids are not rejected. " +
+			"`substreams tools extract-proto --sql` writes the annotated proto to start from, and --proto-file-override points the sink at it")
```
maoueh, line 476: Point to docs.substreams.dev for Relational Mappins to reference that.

cmd/substreams/sink_sql_common.go
```diff
@@ -453,25 +684,210 @@ func newSinkSetupE(driver string) func(*cobra.Command, []string) error {
...
+// registered on the same command and half of them are inert for any given run. They used
+// to be inert silently; a flag typed here means the operator expects the sink to be doing
+// something it will not do, and a warning in a log that scrolls past is not how they
+// should find that out.
+func rejectFromProtoFlags(cmd *cobra.Command) error {
+	names := append(append([]string{}, fromProtoSchemaFlagNames...), fromProtoRunFlagNames...)
+
+	return rejectFlags(cmd, names, "from-proto",
+		"This one outputs DatabaseChanges, where rows are written from the module's own database changes "+
+			"and the schema is the 'schema.sql' bundled in the manifest")
```
maoueh: Each rejectFlags `because` should have a link to some docs reference.

cmd/substreams/sink_sql_common.go
```diff
@@ -453,25 +684,210 @@ func newSinkSetupE(driver string) func(*cobra.Command, []string) error {
...
+				return err
+			}
+			zlog.Info("constraints created", zap.Duration("duration", time.Since(startAt)))
+
+			return nil
+		}
+
+		if err := dropDatabaseConstraints(database); err != nil {
+			return err
+		}
```
maoueh: A confirmation should be asked just in case IMO

cmd/substreams/sink_sql_common.go
```diff
@@ -511,32 +928,43 @@ func runFromProtoSetup(cmd *cobra.Command, driver, dsnString string, spkg *pbsub
...
 	}
 	if !useProtoOption {
-		useConstraints = false
+		// Without the schema annotations there are no relations to constrain. The index on
+		// _block_number_ is not declared by them either, so it stays.
+		constraints = protosql.DisableAllConstraints().WithBlockNumberIndex(constraints)
+
+		zlog.Warn("the module's output carries no schema.proto annotations, so its tables get no primary keys, no unique constraints and no foreign keys. " +
+			"Queries against them are sequential scans and duplicate ids are not rejected. " +
+			"`substreams tools extract-proto --sql` writes the annotated proto to start from, and --proto-file-override points the sink at it")
```
maoueh, line 747: Same here, revisit those integrity warnings/errors and ensure we have proper backlinking to docs or README or some other material.

cmd/substreams/sink_sql_common.go
```diff
@@ -767,3 +1195,142 @@ func cursorToShortString(in *sink.Cursor) string {
...
+
+func parseByteSize(cmd *cobra.Command, name string) (int64, error) {
+	raw := sflags.MustGetString(cmd, name)
+	parsed, err := humanize.ParseBytes(raw)
+	if err != nil {
+		return 0, fmt.Errorf("invalid --%s %q: %w", name, raw, err)
+	}
+
+	return int64(parsed), nil
+}
```
maoueh, line 1066: Very good candidate for sflags addition too.

cmd/substreams/sink_sql_common.go
```diff
@@ -767,3 +1195,142 @@ func cursorToShortString(in *sink.Cursor) string {
...
+
+// flagChanged reports whether the operator actually typed the flag, as opposed to it
+// carrying its default. Every mode and deprecation check keys off this: a default value
+// must never be read as an instruction.
+func flagChanged(cmd *cobra.Command, name string) bool {
+	flag := cmd.Flags().Lookup(name)
+
+	return flag != nil && flag.Changed
+}
```
maoueh, line 1075: sflags has `Provided` variant on all getter `MustGetBoolProvided` returns two value `T, bool` where second is if the flag.Changed

(Maybe Changed would be a better word in sflags actually).

maoueh (approved):

PR #874, opened 2026-08-11
Expose per-stage squash/readiness on RPC progress messages

maoueh:
## Problem

`ModulesProgress.stages` only carried `completed_ranges`, which counts a store segment as soon as its **partial has been produced** — before tier1 has squashed it into the store.

Consequence for a client: a stage renders as 100% covered while tier1 is still squashing. During squashing there are no running jobs (it happens on tier1, not on a worker) and `processed_blocks` does not advance, so the request looks stuck at 100% with a rate of zero while a substantial fraction of the work is genuinely outstanding. Observed live: a request sat at a stage showing 100% while the overall work counter was at 39%, then progressed to 81%, and only then started streaming.

## What this adds

Three additive fields on `sf.substreams.rpc.v2`:

| field | # | Go accessor | unit |
|---|---|---|---|
| `Stage.ready_up_to_exclusive` | 3 | `GetReadyUpToExclusive()` | block number, exclusive |
| `Stage.squash_wait_segment_count` | 4 | `GetSquashWaitSegmentCount()` | segment count |
| `SessionInit.segment_block_count` | 11 | `GetSegmentBlockCount()` | blocks per segment |

`ready_up_to_exclusive` is the honest "usable up to here" frontier: the lowest contiguous block across the stage's modules, stopping at the last **squashed** segment for stores. `completed_ranges` includes produced-but-unsquashed segments; this does not. That distinction is the whole point.

`segment_block_count` was added because the segment width was not exchanged anywhere, so a client had no way to turn a segment count into blocks. It could only be inferred from `Job.stop_block - Job.start_block`, which is unavailable exactly when it matters, since no job runs while tier1 squashes.

### Semantics worth knowing

- **Map-only stage:** mapper/index output is read straight from the partial exec-out files, so partials count as ready and `ready_up_to_exclusive` tracks `completed_ranges`. `squash_wait_segment_count` is always 0.
- **Stage not started:** reports `max(module initial block, chain first streamable block)` — where the modules *begin*. **0 is a valid value** meaning "nothing usable yet", not a sentinel for "unknown". Clients must not branch on 0.
- **`segment_block_count` is an upper bound multiplier**, not exact: the first and last segment of a run are genuinely narrower when a module's initial block or the request's stop block falls mid-segment. Good for sizing a progress bar, not for reconciling a total.

## Cost

Both stage values already existed server-side from #863, computed for the tier1 progress **log line** only. `UpdateStats` now hoists `stagesProgress(stats)` into a local that feeds both the RPC message and the log, so the segment matrix is still walked **exactly once** per update — no second O(segments × stages) pass.

## Testing

- `TestUpdateStatsReadyUpToExcludesUnsquashed` goes end to end to the `pbsubstreamsrpc.Stage` a client receives: a store stage with a single squashed segment and four unsquashed partials reports full `completed_ranges` coverage `[0, 50)`, `ready_up_to_exclusive` still at 10, and `squash_wait_segment_count` of 4.
- `TestStagesProgressIsIndexedByStage` guards the positional read `UpdateStats` relies on.
- `TestEmitLinearSessionInitCarriesSegmentBlockCount` pins the segment width onto the session message.
- Both new assertions were verified non-vacuous by removing the wiring and confirming they fail before restoring it.
- `go test ./...` clean. `go test -race ./...` fails only in `./test/`; that set was confirmed identical on clean `develop` and is unrelated (those integration tests are flaky under `-race` in both directions).

## Notes for review

- **Codegen:** a plain `buf generate` *downgrades* every `.pb.go`, because the repo was generated with `protoc-gen-go` v1.36.10 while a typical local install is older, and the rawDesc format changed. Generated here with v1.36.10 / `protoc-gen-go-grpc` v1.6.0 pinned in a throwaway GOBIN. The repo has no Makefile or `devel/` script pinning these — worth fixing separately.
- The `rpc/v2` generated diff includes some pre-existing stale-comment drift: doc comments added to the `.proto` earlier were never regenerated. Kept, since reverting only defers it. Drift in `rpc/v3`, `rpc/v4` and `sink/` was reverted to keep the diff contained.
- `metrics/progress_log.go` still emits the tier1 log key as `squash_wait_segments`, which no longer matches the proto field name. That key shipped in #863 and ops may be grepping it, so aligning it is a separate call — happy to do it here if preferred.
- Nothing under `tui/` is touched; a separate in-flight branch consumes these fields.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:5cefb22</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:5cefb22</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:7b044abded44f816def76dfb57086f2911a605d45f1ac19a0972626803e75e8b</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

sduchesneau (approved):
LGTM

PR #875, opened 2026-08-11
Report backprocessing as progress and rates in `substreams run`

maoueh:
## Why

While a request backprocesses, `substreams run` printed a list of block ranges plus two sporadic sections. It answered none of the questions a user actually has: how far along am I, how fast is it going, when does data start flowing, and what is holding it up.

Real output from a run against `hoodi`, ten lines of which three were blank:

```
Connected (trace ID 8bf2e35f5a387311a74c4d08f0678b52)
Progress messages received: 13 (5/sec)

(hit 'm' to switch mode)

Stage 0: erc20_metadata:erc20_stores:erc20:map_events,erc20_metadata:erc20_stores:store_erc20_transfers


stage 0 (0 jobs)                  0  ::  0-3385000

```

Nothing there says the only thing that mattered on that run: every block was already cached, so there was no work to do at all.

## What it looks like now

```
Session  b9515ff339de7705220eafdf5e6705f3
  Module  map_tx_counter_summary · production · 2 stages
  Chain   localhost:55717 · start 15,000 · 10 workers max
  Work    14.9k prepare stores · 200 in range

Backprocessing  █████████▓░░░░░░░░░░  47%  1.7k blk/s  eta 10m32s  4/10 jobs  (1M cached)
  s0            ████████████████████ 100%
  s1            ██████░░░░░░░░░░░░░░  30%  4 jobs, oldest 41s
  out           ██████░░░░░░░░░░░░░░  30%  1,150,000 → 1,500,000

Slowest modules
  s1  map_pools_created  25ms/blk (avg 13ms)  [eth_call (4): 61%]
  s0  store_pools        12ms/blk             [store (3 read/blk, 1 write/blk, 0 deletePrefix/blk): 30%]
```

The session header is printed once and stays in the scrollback — the live view is torn down the moment the first block arrives, so a trace ID rendered by the model would vanish exactly when it becomes useful. Every bar goes through one layout function, so they cannot drift out of column.

Three states, not one skeleton: `nothing to do — 3,385,000 blocks already cached`, the block above, and `complete`.

## Where the numbers come from

No new data on the wire beyond #874; most of it was already being sent and discarded.

- **Percentage** — `processed_blocks + Σ running_jobs[].progress_blocks` over the effective work counts in `SessionInit`. `processed_blocks` alone only moves when a job completes, so the bar stepped; adding in-flight progress makes it continuous, which is what makes the rate stable.
- **Rate and ETA** — a trailing 30s window anchored on *time*, not on a sample count, because the server stretches its progress interval from 500ms to 5s over the life of a request. ETA is suppressed below a 10s span.
- **Stage progress** — `Stage.ready_up_to_exclusive` from #874, never `completed_ranges`. See below.
- **Module cost** — recent (window delta) alongside the lifetime average, floor 10ms/blk, ranked across all stages with a stage tag.

### Squashing was invisible, and `completed_ranges` lies

`completed_ranges` counts a store segment as soon as its partial has been **produced**, before tier1 has squashed it. Since squashing runs on tier1 rather than on a worker, it schedules no job and advances no `processed_blocks` — so a stage read 100% with zero jobs and a rate decaying to zero, and the request looked stalled. It is now named:

```
s0  ▓░░░░░░░░░░░░░░░░░░░   1%  10 jobs, oldest 12s  squashing 1 segment
```

Reading the squashed frontier fixed a second bug in the same place. `ready_up_to_exclusive` is the lowest *contiguous* block across a stage's modules, so a stage with a gap in its produced ranges no longer counts the ranges beyond the gap — in the golden fixture that took a stage from a bogus 50% to an honest 30%, now agreeing with the `out` row — and a stage that has not started reports where it begins instead of having no anchor at all.

An earlier version of this branch inferred "parallel finished" from all stages at 100% plus no running jobs. That condition is true *during squashing*: on a real run it fired at 39% of the work. It was reverted, and the reasoning is recorded in a comment so nobody re-derives it.

## Failure output

Sections are separated, the progress block is closed rather than trailing off at `starting…`, and the most common first-run failure names its own fix:

```
Backprocessing  aborted

📊 Usage Report: no data received

Error: --limit-processed-blocks is set to 100, but this request needs to process 4,100 blocks (3,900 of them to prepare the stores): raise it with --limit-processed-blocks 5000, or remove the guard entirely with --limit-processed-blocks 0
```

The figures come from what the client was told at session init, not from parsing the server's message.

## Fixed along the way

- **Ctrl-C needed two presses.** The live view puts the terminal in raw mode, so the first press arrived as a key event and never as SIGINT: the UI quit while the request kept streaming, and the second press only worked because the first had released the terminal.
- **A header line that never rendered.** `Backprocessing history up to requested target block` was guarded on `.Request`, set only by `TUI.SetRequest` — which had no callers anywhere. Removed, along with the `m` key, whose two rendering modes no longer exist.
- **TTY users saw less than piped users.** The head block, stage count and cached-blocks summary were printed only on the non-TTY path.

## Supporting changes

- **`internal/formatx`** collects the block, count, rate, duration and byte formatting that `cmd/substreams/estimate.go` and the TUI had each grown a copy of. Writing its tests immediately caught two bugs in code this branch had already produced: `59m59.5s` rendering as `60m00s` and `999,999` as `1000k`, both rounding applied after the unit was chosen.
- **`substreams tools devenv`** boots a local stack — dummy blockchain container plus tier1 and tier2 built from this tree — so a real backprocess can be watched without a remote endpoint. It waits for the merger to catch up before starting tier1, since tier1 bootstraps its hub from merged blocks and otherwise never becomes ready. The end-to-end tests share the same setup code, and replacing their in-container port probe with a host-side dial took the container test from 150s to 12.8s. It lives here rather than in firehose-core because firehose-core depends on substreams, so a command there could only run the version it pins, never the working tree.
- **`Sinker.PrintStats`** collapses to one line when a request produced nothing, instead of a header over three zeroed counters. Also affects `sink webhook` and `sink noop`.
- The `text/template` view is gone in favour of direct string building — the template was the source of the dead-code branch above, and a typo in it degraded to a rendered error string rather than a compile failure.

## Testing

Golden tests over `model.View()` with an injected clock cover all three states, the stage collapse and drop paths, squashing, leaf-name collisions, the cost floor and a narrow terminal, plus a check that every bar starts in the same column. Ctrl-C, the limit-error explanation and the formatting helpers have their own unit tests.

Verified end to end against a local stack: a cold backprocess, a fully cached re-run, the limit failure, and a real `\x03` fed into a pty for the interrupt.

`go build ./...`, `go vet` and the test suites pass in both modules; `go.mod` is untouched (testcontainers was already a direct dependency).

## Note for review

The tier1 progress log still emits `squash_wait_segments` while the proto field is `squash_wait_segment_count`. That key shipped in #863 and ops may be grepping it, so aligning the two is left as a standalone call.


tools/devenv/stack.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+
+	t1app := app.NewTier1(logger, conf, &app.Tier1Modules{
+		HeadTimeDriftMetric:   metricset.NewHeadTimeDrift(prefix),
+		HeadBlockNumberMetric: metricset.NewHeadBlockNumber(prefix),
+		Authenticator:         auth,
+		SessionPool:           sessionPool,
+	})
+	go func() {
+		if err := t1app.Run(); err != nil {
```
UlysseCorbeil: the error from Run() is logged but if Run fails we get a "timeout after 2m" without a specific indication of what happened, could be worth returning in waitReady

tools/devenv/cmd.go
```diff
@@ -0,0 +1,147 @@
...
+		Burst:           burst,
+		BlockRate:       sflags.MustGetInt(cmd, "block-rate"),
+		ExtraReaderArgs: sflags.MustGetString(cmd, "reader-args"),
+	})
+	if err != nil {
+		return fmt.Errorf("start dummy blockchain: %w", err)
+	}
+	defer func() {
+		// A fresh context: cmd's is already cancelled by the time this runs.
+		if err := container.Terminate(cmd.Root().Context()); err != nil {
```
UlysseCorbeil: so if its already cancelled Terminate returns without stopping the container ? Could use a WithTimeout

UlysseCorbeil (approved):

sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:5b48541</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:5b48541</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:2bb14310a97149dac01a96331f24ed8559d03a17086e36645b54095c91f740b5</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

PR #880, opened 2026-08-13
Add a `context::clock` WASM intrinsic

maoueh:
## What

Adds a `context` WASM host module with a single intrinsic, `context::clock`, which writes the current block clock into guest memory as an encoded `sf.substreams.v1.Clock`:

```rust
#[link(wasm_import_module = "context")]
extern "C" {
    fn clock(output_ptr: *mut u8);
}
```

It follows the same calling convention as the `state` getters — the host allocates the payload on the guest heap and writes a `{ptr, len}` pair at `output_ptr`. It returns nothing: unlike `state::get_at` there is no "not found" case, the clock always exists. A marshalling failure is raised as non-deterministic, so a transient failure is never cached into the store.

Implemented for **wasmtime** (the default runtime) and **wazero** (which also covers `wasip1/tinygo-v1`). The JavaScript/v8 runtime is not covered.

`context` also joins `env`, `state` and `logger` as a namespace WASM extensions cannot register into, keeping it available for whatever belongs in it next.

## Why

Until now the only way for a module to reach the clock was declaring `source: sf.substreams.v1.Clock` as a module input, which forces the dependency into the module graph even when the clock is needed only occasionally. An intrinsic is callable at any point during execution and costs nothing when unused.

## Scope note: network id was dropped

The branch name says `network-id` because it started as "expose the deployment's network id to modules", with the clock added alongside. **The network id half is not in this PR.** No compelling use case turned up for it, and a `networks:` block in the manifest already parameterizes a package per network, which covers the same ground.

The investigation is worth recording, since it also shows the network id would have been awkward to deliver:

- firehose-core already resolves the right value — `--advertise-chain-name` is looked up in the networks registry, canonicalized from aliases, and cross-checked against the genesis block id. It just never reaches the substreams tier1 app.
- That flag is optional. Without it the chain is inferred from the genesis block during `InfoServer.Init`, which needs the block stores to be reachable and so resolves well after tier1 config is built.
- Most large deployments rely on that inference: in sf-operator, 21 of 46 environments set `advertise_chain_name`, and the ones that don't include eth-mainnet, eth-hoodi, eth-sepolia, arb-one, eth-bsc-mainnet and eth-polygon-mainnet.

So any tier1-side network id would have started out unset on the busiest deployments, ruling out failing fast on a missing value and leaving a degraded mode to design around. That's the constraint to solve first if the feature comes back.

## Testing

- Unit test: `Call.DoClock` round-trips through protobuf.
- End to end: a new `assert_context_intrinsics_0` module in the `complex_substreams` testdata package asserts, inside the WASM, that the intrinsic matches the Clock it received as a module input. A mismatch panics in the module, so a `true` output for every block is the proof. Exercised over blocks 0-10 via the tier2 integration test, on both wasmtime and wazero.

The committed `.spkg` is rebuilt, which shifts every module hash in that package.

## Notes for reviewers

- The first commit is a prerequisite, not part of the feature: the `complex_substreams` testdata crate could not compile at all, because its generated `pb/mod.rs` declared four modules whose `.rs` files are absent from the directory. That had to be fixed before the package could be rebuilt.
- Ergonomic Rust bindings (`substreams::context::clock()`) belong in substreams-rs and will follow separately. Until then modules declare the extern by hand, as the test fixture does.
- Purely additive: no config, no proto change, no behaviour change for existing modules.


UlysseCorbeil (approved):
Solid, lgtm

sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:88cdec9</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:88cdec9</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:f9a7ef56d2ee11643dc45fc0fec172cd109bb3743cd2f9b0876fcf3cbd2df77d</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
</table>
</details>

<table></table>

sduchesneau (approved):
LGTM.
Let's not forget substreams-rs updates to add the helper once it's shipped.

PR #898, opened 2026-08-21
Sink SQL: Report what the spool is doing, and stop losing it on shutdown

sduchesneau:
## Why

A `sink postgres|clickhouse` backfill in Relational Mappings Mode buffers rows through a local spool. The statistics panel reported the *gap* between download and database, but nothing about the write path closing it — so "the database is 400k blocks behind" could mean the database is slow or the stream burst, and there was no way to tell which. Separately, interrupting a run — the normal way a backfill ends — silently threw away the open spool segment.

## What

**1. Wait for the sink to shut down before exiting.**

`cli.Application.WaitForTermination(zlog, 0, 0)` treats a zero graceful delay as *do not wait* — unlike `BlockUntilTerminated`, where zero means *wait forever*. On SIGINT the process returned through `time.After(0)` while `Sinker.Run` was still being cancelled, so its defers never ran: the final statistics were lost, and so was `closeDatabase()`, which is what seals the open spool segment. Every block in that segment was re-streamed on the next start. Every other call site in the repo already passed 30s; this one was the outlier.

**2. Account for what the applier commits.**

The spool records segments, blocks, rows, bytes, per-commit duration, seal reasons, quota wait, and how much of the applier goroutine's wall clock is spent working rather than waiting on the queue. That last ratio is the one that says whether the database or the stream is the limit.

ClickHouse counts bytes on its own socket via a `ch.Options.Dialer` wrapper, kept distinct from the spooled segment size — they differ by the columnar encoding and LZ4, often several-fold, and conflating them misleads anyone sizing a link. `pgx` owns its connection so the PostgreSQL path can't answer this, and reports nothing rather than a zero.

**3. Fold the account into `Progress` and fix what the panel claimed.**

`Progress` already owned the buffered-blocks and quota numbers, so the spool's account lives there rather than in a parallel tracker being handed the same values from the same call and rendering them twice.

- `Flush duration` is fed by what the applier committed. With a spool, `Database.Flush()` returns before anything reaches the server, so it was reporting a structural `0.00ms` — a database that looks instant for the whole backfill.
- `Block Insert Duration` was added once per batch while the rows around it were added per block: three averages in one column, one of them in a different unit that grew with `--decode-batch-size`. Now per block.
- Durations render per 100 blocks. A per-block mean lands in fractions of a microsecond, and a column of `0.02ms` / `0.08ms` can't be weighed against anything. The stored value is unchanged — the live-flush heuristic still compares it against a block time.
- `Entities Insert Duration` → `Message Walk Duration` (it was always the walk, never an insert); `Block Insert Duration` → `Spool Write Duration` while a spool is open.
- `fallingBehind` compared buffered bytes against the quota, and the quota outlives the spool. After the switch to direct inserts nothing is buffered, so that branch returned false forever and the sink could never report falling behind again.

**4. Announce recovery while it runs.**

Replaying the segments a killed backfill left behind happens inside `Open()`, before anything else logs, and takes as long as the COPYs take. It reported only after finishing, so it read as a hang at startup — which is exactly how this PR started.

## Panel

```
   Wait Duration Between Blocks {"per_100_blocks": "2.00ms", "recent_per_100_blocks": "2.00ms"}
      Block Processing Duration {"per_100_blocks": "0.1s",   "recent_per_100_blocks": "0.1s"}
         Unmarshalling Duration {"per_100_blocks": "1.00ms", "recent_per_100_blocks": "1.00ms"}
           Spool Write Duration {"per_100_blocks": "6.00ms", "recent_per_100_blocks": "6.00ms"}
          Message Walk Duration {"per_100_blocks": "8.00ms", "recent_per_100_blocks": "8.00ms"}
                 Flush duration {"per_100_blocks": "0.12s",  "recent_per_100_blocks": "0.12s"}
WARN database is falling behind... {"buffered_on_disk": "8.0 GiB", "quota_used": "100.0%"}
                  Spool applier {"segments": 133, "rate": "18.0 seg/min", "blocks_per_seg": 2703, "apply_duration": "3.11s", "queue_depth": 121}
               Spool throughput {"rows": "63,999/s (23,147,888 total)", "spool_rate": "36 MiB/s", "spooled_bytes": "7.0 GiB"}
                 Spool pressure {"applier_busy": "100.0%", "quota_wait": "4m 15s", "segment_target": "74 MiB", ...}
```

## Interface change

`Database.BufferStats()` returns `(sql.WriteStats, bool)` instead of a positional `(int64, int64, uint64, bool)`. `WriteStats` embeds `spool.Stats` and adds the one field only the driver can answer, so the spool never carries a number it doesn't produce. Two implementers, both delegating.

## Notes for review

- Rates are differenced between logging ticks, not averaged over the run — a lifetime average reports a healthy rate through a ten-minute stall.
- The spool panel has three states: never spooled (absent), spooling (full detail), closed at chain head (totals only). The third matters because reaching HEAD closes the spool mid-run, and rates over a stopped applier read as one that has stalled.
- Every commit builds independently.
- `Block Processing Duration` is still contaminated: `awaitQuota` blocks *inside* the processing timer, so backpressure inflates "processing" and deflates "wait between blocks". Left alone here — the fix is to split blocked-on-database time out of `TotalProcessingDuration`, which is a behaviour change worth its own PR.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:49b1181</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:49b1181</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:82c8f5150433764b5ffc9cd947bcd51d17f9b86e0d5cdce7f515d1a9d9ec4aa5</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 4 low: 4 </td></tr>
</table>
</details>

<table></table>

sduchesneau:
(from Claude, doing review and addressing it) 

Pushed `644630c3` addressing the review. The two most serious findings were real — I verified both in the code before fixing rather than taking them on trust.

**`applier_busy` inverted under load.** `applierBusy`/`applierIdle` only accrued when an activity *ended*, so a segment outlasting the 30s logging interval contributed to neither side: `0/1ns` → `applier_busy: 0.0%`, with `rate: 0.0 seg/min` and `apply_duration` dropped by the `segments > 0` guard. The panel read "applier idle, stream is the limit" at precisely the moment the database was the limit. `quota_wait` had the same shape — deferred until `awaitQuota` returned, so a single long hold reported `0s` throughout. Both now stamp an in-flight marker that `Stats()` folds in.

**Counter ordering dropped samples.** `appliedSegments` was incremented before blocks/rows/duration, so a snapshot in that window saw `segments>0, blocks==0`. `RecordBuffered` gated the `FlushDuration.Add` on `blocks > 0` but advanced `lastCommit` unconditionally, discarding that segment's cost permanently. The segment is now counted *after* the totals it divides, and `lastCommit` only advances when a sample was actually taken.

**Totals stopped short of the final drain.** Closing the spool drains everything queued — up to the whole 8GiB budget — and the last snapshot was taken before it. Both drivers now capture `Stats()` after `Close` and keep reporting it with `enabled=false`; `closeDatabase` takes one more reading so the run's final panel includes it.

**Bounded shutdown drain.** `Close` ended in an unbounded `waitGroup.Wait()`, so the 45s in the first commit was a guess. It now respects its context and logs what it left for recovery — sealed segments are durable, so an abandoned drain is safe. `SwitchToDirectInserts` deliberately keeps an *unbounded* drain via `context.WithoutCancel`: the applier advances the stored cursor as it goes, and abandoning it there would rewind the cursor behind blocks the direct path had already written. That distinction wasn't in the review and is the reason this isn't a blanket change.

Also: dead `insertDuration` plumbing removed from `decoder.apply`; `bufferedBytes` loaded once so the warning can't claim a buffer the same line doesn't show; sealed/open byte reads moved under one lock; negative guard on the spool's byte formatting; `elapsed` guard moved inside the spooling case and raised to 1s so the final `LogStats` can't print a nonsense rate.

**Tests.** The recovery test now asserts log *ordering* (it would previously have passed with the announcement moved to the end — fair hit). Added panel arithmetic assertions, a two-tick delta test, an in-flight `applier_busy` guard against a blocking applier, and a `Close`-respects-deadline test. Writing the arithmetic test caught one of my own wrong assumptions: I expected `spooled_bytes` to be a delta, and it's cumulative by design.

Not addressed, deliberately: the `Average` slice data race between `Add` and the 30s ticker is pre-existing, though this PR does raise exposure; and `awaitQuota` blocking inside the processing timer still contaminates `Block Processing Duration`. Both are separate PRs.

`go build`, `go vet`, `go test ./sink/... ./cmd/...` and `go test -race ./sink/sql/db_proto/...` all clean. All four commits build independently.

sduchesneau:
@copilot resolve the merge conflicts in this pull request

billettc (approved):

PR #901, opened 2026-08-25
Support pruned store snapshots (thin store)

sduchesneau:
Lets operators delete intermediate store snapshots (fullKV files) of long-lived substreams to save disk space, without breaking requests. Tier1 now rebuilds stores from the last snapshot that still exists instead of assuming every earlier one is present.

Companion PR: streamingfast/firehose-core#225 — `firecore tools substreams prune-states` (the pruning tool) and `devel/thinstore` (the local test bench used below).

### Tier1 — `Stages.FetchCachesState` (`orchestrator/stage/fetchstorage.go`), once per request

1. **Mapper outputs** (production mode only). Lists the output module's execout files over the `WriteExecOut` range and marks each existing segment `Completed`. If every segment is there and no linear pipeline follows, the store stages are all marked `NoOp` and nothing else is listed. *Assumes:* nothing about earlier files; a missing execout segment is simply regenerated. Intermediate mappers' outputs are never checked here (tier2 decides, see below).

2. **Target segment** (`storesTargetSegment`). The last store segment that must be complete before any job can run: the segment right before the first mapper segment still `Pending`, or the last store segment when the linear pipeline needs the stores at the handoff block (dev mode, or all mapper outputs present with a live tail).

3. **Snapshot discovery** (`scanStoreSnapshots`). Lists fullKV/partial files in windows walking **backwards** from the target: 50 segments, then ×4 each time (`storesWalkWindowSegments`, `storesWalkWindowGrowth`), stopping as soon as a resume point is found or the lowest module initial block is reached. The first window also covers everything above the target so snapshots/partials left by an earlier run are reused. Cost is `O(log distance-to-nearest-snapshot)` bounded listings, never a listing of the whole history. *Previously:* one 50-segment window, then a full-history fallback listing whenever a store had no snapshot in that window — that fallback would fire on every request once snapshots are pruned.

4. **Resume segment** (`resumeSegment`). The highest segment `≤ target` at which **every** store module of **every** store stage (that has started by then) has a fullKV. Stores with different holes therefore fall back to their last *common* boundary, because a tier2 job at segment `j` loads all lower-stage stores at the same block `j*interval`. `-1` means "nothing usable, start from the initial block".

5. **Marking units.** Segments `< resume` on store stages → `NoOp` (nothing will ever read those files, present or not). Segments `≥ resume` → `Completed` **only for fullKVs actually seen** (all modules of the stage), `PartialPresent` for partials seen above the resume point. *Previously:* the first fullKV seen back-filled every earlier segment as `Completed` ("a fullKV at x implies all fullKVs < x exist"), which is exactly the assumption pruning breaks; the assumption is gone.

6. **Scheduling dependency** (`Stages.dependenciesCompleted`). A job at `(stage N, segment j)` now requires, for every lower stage, unit `(s, j)` done **and** unit `(s, j-1)` done — the latter guarantees the fullKV at `j*interval` exists (either seen on disk or just written by the squasher). *Previously:* only `(s, j)`, which was enough under the old assumption. `MoveSegmentCompletedForward` walks over `NoOp` so the squasher's "last completed" cursor skips the pruned prefix.

7. **Squash** (`orchestrator/stage/squash.go`, unchanged). `getStore(rng.StartBlock)` loads the fullKV at the segment's start: at `resume+1` that is the surviving snapshot; afterwards it is the one the previous squash just wrote. Every squash still writes a fullKV at each interval boundary, so pruned holes get refilled on the way. `getPartialOrFullKV` still adopts an existing fullKV at the segment's end if one survived above the resume point.

8. **Linear pipeline start** (`Stages.FinalStoreMap`) loads every store's fullKV at the handoff block — guaranteed by 4–7 to have been written.

### Tier2 — per job (`service/tier2.go` → `pipeline.setupSubrequestStores`)

- Before executing, tier2 checks per module whether its output for this segment already exists: execout file for mappers/indexes; fullKV at the segment's end, or a partial for the segment, for stores (`ExistsFullKV`/`ExistsPartialKV`, one HEAD each). Modules already done are not re-executed; if nothing is left, the job returns immediately. Unchanged.
- For lower-stage stores it then loads the fullKV at the job's **start block** (`NewCompleteFileInfo(init, ResolvedStartBlockNum)`). *Assumes* it exists — tier2 cannot rebuild a lower stage. This is what tier1's step 6 guarantees. A missing file now fails with `snapshot "…" does not exist (it may have been pruned)` instead of a bare `not found`.
- Stores of the job's own stage are computed from scratch for the segment and written as a partial; the last stage also writes fullKVs at boundaries where none exist (`Stores.saveStoresSnapshots` skips existing ones).

### Scheduler deadlock found by the chaos fuzzer (fixed here)

`markShadowedUnits` shadowed a unit whenever the unit of the stage above was `Pending`, `Scheduled`, `Shadowed` **or `Merging`**. Only a job of the stage above ever un-shadows a unit; a unit above that is `Merging` got its partial from disk (left by an interrupted run), so no job ever came back and the shadowed unit stayed `Shadowed` forever — the request sat in `parallel_processing` with idle workers. Pruned snapshots make the "upper stage has a partial, lower stage still pending" pattern common, so `Merging` is no longer a shadowing trigger (`TestShadowNotUnderMergeFromDisk`).

### Output file deleted after its job ran (fixed here)

`ExecOutWalker` retried a missing output file forever (`MsgFileNotPresent` → re-download, no bound) when the file had been deleted after the job wrote it — the request sat with every stage complete and the stream stopped. After 30s of a file missing for a segment whose map unit is done, the unit goes back to `Pending` (`Stages.ReprocessMapSegment`, which also rewinds `nextJobCursor`) and the job is scheduled again.

### Snapshot pruned under a running request (fixed here)

A job of stage N+1 loads the stage-N fullKV at its start block. When that file disappears while the request runs (the fuzzer's chaos mode, or an operator pruning a range a long request is still rebuilding), tier2 answered `Internal` and tier1 retried the job forever (`RetryContext(ctx, MaxUint64)`): stalled request, idle workers, "stage N lost 8 job(s)" hints. Now `FullKV.Load` wraps `store.ErrSnapshotMissing`, tier2 maps it to `FailedPrecondition` (`… does not exist (it may have been pruned); restart the request`) and the worker treats it as terminal, so the request fails at once and a restart computes a fresh resume point.

### Workers leaked when a request ends with jobs in flight (fixed here)

`SessionWorkerPool` returned still-borrowed workers asynchronously on context cancel, usually after tier1 had already released the request's session. `LocalSessionPool.ReleaseWorker` then finds no session for the worker and never decrements the global/organization counters. Every request ending on an error while jobs run (which the pruned-snapshot failure above now does promptly) leaked its in-flight workers; after enough of them every later request stalled in `streaming_first_segment` with "no worker was free" and 0 running. `ParallelProcessor.Run` now calls `WorkerPool.ReleaseAll()` synchronously before the request releases its session (`TestSessionWorkerPoolReleaseAllBeforeSessionRelease`).

### Caveats

- A snapshot deleted **while** a request still needs it fails that request with `FailedPrecondition … restart the request` (no wrong output, no hang, no endless retry). `prune-states --keep-recent` exists to keep the head's neighbourhood intact; pruning far behind the head is safe.
- Stores whose surviving snapshots don't line up rebuild from the last common boundary, so keep the same `--keep-every` for all modules of a substreams.
- A store with no snapshot at all still walks its listing down to its initial block (bounded windows, but all of them). That is the "first ever run" case only.
- No index/partial/execout semantics changed: partials are still transient, index and intermediate-mapper outputs are still regenerated when missing.

## Tests

The bench now lives here as the opt-in `tests_e2e` suite `TestThinstore` (`SUBSTREAMS_E2E_THINSTORE=true`): dummy chain container, in-process tier1/tier2 with 100-block segments, the `tests_e2e/thinstore` Clock-only package, baseline over 15000 blocks, the scenarios below, then seeded fuzz iterations (`..._FUZZ_ITERATIONS`, `..._FUZZ_SEED`, `..._FUZZ_CHAOS`) with hang detection (goroutine dump of the in-process tier1).

Unit (`orchestrator/stage`): existing `FetchCachesState` cases re-baselined (no back-fill), plus pruned single store, two stores with different holes, two stages with different holes, no snapshot at all (window walk-back to genesis), and `dependenciesCompleted` requiring the previous segment of every lower stage.

Local bench (firehose-core `devel/thinstore`): dummy chain, 30k merged blocks, offline tier1/tier2, snapshots every 100 blocks, 4-stage Clock-only package (`index_parity, map_a, store_count, store_a_max` / `store_sum, store_a_last, map_b` / `store_c, map_c[filtered]` / `map_out`). Baseline `[0,30000)` in production mode, then `prune-states --keep-every 1000` plus extra per-module holes in snapshots, outputs (intermediate and final) and index files. Every scenario compares each block's output with the baseline and scans the tier1 log for pruned-file loads. All passed:

- `outputs_present_near_head [28000,29000)` prod — final outputs kept near head while older ones are deleted; served from cache.
- `outputs_partially_present [2300,2700)` prod — final outputs missing for the first part of the range and present after.
- `outputs_missing_then_present [24200,24500)` prod — a 300-block output hole ending inside the range.
- `outputs_present_then_missing [26300,26800)` prod — outputs present at the start, deleted from mid-range.
- `genesis_all_missing [0,500)` prod — outputs deleted from genesis, resume from scratch (`-1`).
- `index_hole_and_outputs_missing [1100,1300)` prod — index files and map_a/map_out outputs all missing.
- `index_hole_outputs_present [14000,14500)` prod — index deleted but final outputs present; nothing rebuilt.
- `stage2_holes [3200,3500)` prod — stage-2 store missing snapshots 3000/4000; resume at 2000 for all stores.
- `stage3_big_hole [6000,6500)` prod — stage-3 store missing 5000…9000, map_c and map_out outputs deleted; resume at 3400.
- `stage3_big_hole_to_kept [7300,8200)` prod — same hole, range ends past a kept boundary.
- `stage1_and_stage2_hole_at_20000 [19500,20500)` prod — stage-1 and stage-2 stores both missing 20000.
- `stage2_a_last_hole [15000,15300)` prod — `store_a_last` missing 15000 with map_b outputs deleted around it.
- `upper_stages_missing [26600,26900)` prod — map_c and map_out outputs deleted, stores intact; resume = target.
- `far_range [10500,11000)` prod — stage-1 store missing 10000; resume at 8100.
- `tail_all_missing [29000,30000)` prod — every output deleted up to the last block.
- `dev_stage3_big_hole [6100,6150)`, `dev_stage2_a_last_hole [15050,15100)`, `dev_tail [29950,30000)` — dev mode over the same holes.
- `full_range [0,30000)` prod — whole range again over every hole at once, reproduces the baseline; disk back to 1500 snapshots.
- `rerun_stage3_big_hole [6000,6500)` prod — second pass over a rebuilt range does not resume from scratch.

Fuzz (`devel/thinstore/fuzz.py`): per iteration a random deletion plan per module and file kind (random holes, keep-every-N, contiguous spans, everything below a block, everything but a few, everything, boundaries chosen so stores share no common kept snapshot) then random prod/dev queries biased towards hole edges (±1/±100 blocks), tiny, long and tail ranges, 3 concurrent:

- seed 1: 6 iterations × 5 queries, sequential — 30/30 pass.
- seed 2: 12 iterations × 6 queries, 3 concurrent — 72/72 pass.
- seed 4: 10 iterations × 6 queries, 3 concurrent — 60/60 pass.
- seed 3: same with `--chaos` (random files deleted continuously while requests run) — no wrong output, no panic, no hang; requests that lost a file mid-flight failed with the pruned-snapshot error.

- seed 5 chaos (first run): one dev-mode request hung for hours in `parallel_processing` — the shadowing deadlock above. Fuzzer now kills and reports a query after `--timeout` seconds.

- seed 3 chaos (fixed-shadowing run): `prod [11799,12799)` hung on a mapper output deleted after its job completed — the walker retry above; fixed, chaos seeds re-run afterwards.

- seed 5 chaos (walker-fix run): `prod [5205,16053)` stalled on stage-2 jobs retrying a stage-1 snapshot deleted after squash — the endless-retry case above; fixed, chaos seeds re-run.

Final campaign on the branch head (`run-all.sh`, all fixes in): scenarios 21/21; seeds 1, 2, 4: 162/162; chaos seeds 3, 5, 6: 108 passed, 36 clean casualties (requests that lost a file mid-flight and failed with the pruned-snapshot error), 0 wrong outputs, 0 hangs.

`TestThinstore` on the branch head: default mode (baseline, 19 scenarios, 3 fuzz iterations × 5 queries) passes in ~3 minutes; with `SUBSTREAMS_E2E_THINSTORE_FUZZ_CHAOS=true` it passes in ~9 minutes with every lost-file request failing cleanly, no hang.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:0920d29</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:0920d29</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:f7ea55c919fca0f35cecfefd03182fa1d09674d08afbe2f91b0779702fd35df0</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>124 MB</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 8 low: 7 </td></tr>
</table>
</details>

<table></table>

orchestrator/stage/fetchstorage_test.go
```diff
@@ -146,13 +146,13 @@ func TestFetchOutputMapperState(t *testing.T) {
 					),
 				},
 			},
-			wantStates: `S:CCC
```
maoueh, line 149: Why this one changed without any other changes?

storage/store/full_kv.go
```diff
@@ -172,6 +172,9 @@ func (s *FullKV) Load(ctx context.Context, file *FileInfo) error {
 
 	reader, err := loadStoreStream(ctx, s.objStore, file.Filename)
 	if err != nil {
+		if errors.Is(err, dstore.ErrNotFound) {
+			return fmt.Errorf("load store stream: snapshot %q does not exist (it may have been pruned): %w", file.Filename, err)
+		}
```
maoueh, line 182: Do we need a specialized error here?

maoueh (approved):

orchestrator/stage/fetchstorage_test.go
sduchesneau: We start at 30, we don't need to check if 10 and 20 are completed (C), new feature just checks that 30 is there, marks the others as noop

sduchesneau: (and I know that we start at 30 because the mapper outputs exist up to 30)

PR #902, opened 2026-08-25
refactor(service): use slices.SortFunc for clock ordering

daixiheguu:
Replace reflection-based `sort.Slice` with typed `slices.SortFunc` when ordering block clocks in the service utilities.

The same comparison key and ascending ordering are retained. The change removes reflection and gives the comparator concrete clock types.


maoueh:
Already done elsewhere

PR #906, opened 2026-08-26
Add rolling-window operation mode to tier1

sduchesneau:
Adds a rolling-window mode for tier1 servers, for chains that only keep a rolling window of blocks. Set from firecore through a flag.

## Config surface

`app.Tier1Config.RollingWindowMode` (bool), plumbed through `service.WithRollingWindowMode()` into `RuntimeConfig.RollingWindowMode`.

## Behavior in rolling-window mode

**Stores are disabled.** A request whose used-module graph contains any store module is refused with `InvalidArgument: store modules are not supported on this endpoint: <names>`. A rolling window cannot be walked back far enough to build store state.

**First streamable block validation is relaxed.** A module whose initial block sits below the first streamable block is clamped to it instead of being rejected — the first streamable block moves forward as older blocks are pruned, so it is a moving lower bound, not a fixed chain property. Only an explicit request for blocks below the first streamable block is still refused (that check already existed and is unchanged); `start_block_num == 0` still resolves to the first streamable block.

The lowest-init-block computations already clamped to the first streamable block, so the request plan and its segments still start at the window, not at 0.

## Tier2

Tier2 now always builds its execution graph with the relaxed check. Tier1 dictates the ranges it sends to tier2 and those are always within the streamable window, so the validation had nothing left to protect. No new field on `ProcessRangeRequest`.

Default-mode tier1 keeps the hard error.

## Tests

- `pipeline/exec/graph_firststreamable_test.go` — strict error preserved by default, clamping under the option.
- `service/tier1_rolling_window_test.go` — option wiring on both tiers, store-refusal message.

`go test ./...` passes.

https://claude.ai/code/session_01Tr9RDBAjU4WSiqh4K4z5q2


sduchesneau:
closed for now, we have users with stores on this endpoint so unsure if this is the way to go

PR #908, opened 2026-08-28
Add hosted-store dummy modules to the e2e dummy package

maoueh:
Adds three Substreams modules to `tests_e2e/dummy` so a staging environment can be exercised against both Hosted Store flavors, the way the existing dummy modules cover the ClickHouse/Postgres sinks.

`map_hosted_store_feed` emits `SinkEntries` and is what you give a **Substreams Feed** store as its output module — it writes a `block:<height>` key that is written once, plus a `latest` key overwritten every block, so a consumer can tell a live store from a stalled one. It ships packed as `e2e-v0.3.0.spkg`. The two readers, `map_query_substreams_feed_store` and `map_query_remote_feed_store`, emit `test.output.StoreQuery`: the block, and per key the response code name, the Any's type URL, and the decoded value — so a miss, a Type URL mismatch and a decode failure stay distinguishable. For the **Remote Feed** flavor, `hosted-store.sh` seeds two fixed keys and flips `SetReady` over gRPC using `buf curl` against the BSR-published schema, which needs no vendored protos and no new Go code.

Each reader gets its own manifest — `substreams.substreams-feed.yaml` and `substreams.remote-feed.yaml` — rather than living in `substreams.yaml`. A `foundational-store` identifier is expanded when the manifest is *read*, and `pack` freezes the result into the module hashes, so a deployment id known only after the store exists cannot live in a committed spkg. Splitting them also means each manifest needs only its own store id exported, since every module in a file is expanded on read.

The dummy package's descriptor sets are now pinned to a buf commit. Unpinned sets resolve to latest, which disables the `substreams build` protobuf cache and deletes `.last_generated_hash` on every run.

The `substreams` crate goes 0.6.2 → 0.7.6, which is where `store::FoundationalStore` and the `foundational_store::model::v2` types first appear. `prost` stays at 0.13.3 and the committed `e2e-v0.1.0.spkg` / `e2e-v0.2.0.spkg` are untouched, so the existing e2e tests are unaffected.

Two mismatches with `docs/how-to-guides/hosted-services/hosted-store-remote-feed.md` turned up and are **not** fixed here: `Feed.Delete` does not exist in the schema published on the BSR (only `Set` and `SetReady`), and `Feed.Set` takes a `SetRequest` wrapping `SinkEntries`, not a bare `SinkEntries` as the guide's payload example shows.


UlysseCorbeil (approved):

sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/substreams:cd893d7</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/substreams:cd893d7</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:0c5b7520769e881513413b0bb078e4f5cd7b0f1e756f515d534a36bec8aa2e05</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>124 MB</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 19 low: 7 </td></tr>
</table>
</details>

<table></table>

PR #911, opened 2026-08-28
add ch annotated dummy package as a sibling manifest

UlysseCorbeil:


tests_e2e/dummy/proto/sf/substreams/sink/sql/schema/v1/schema.proto
```diff
@@ -0,0 +1,225 @@
+syntax = "proto3";
```
maoueh: Can we check to avoid having this committed. This is literally a copy of the original source. A buf reference might be better ... but this buf is no on Substreams directly, so most probably it should now be part of Substreams "system protobufs".

maoueh (approved):

tests_e2e/dummy/substreams.clickhouse.yaml
```diff
@@ -0,0 +1,44 @@
+# ClickHouse sink fixture. Same events as map_events in substreams.yaml, emitted
```
maoueh: Let's add annotation directly to proto, no need for a new manifest. We can re-use everything existing + bump to latest + avoid a different .spkg just for this.

The **tests_e2e/dummy/e2e-clickhouse-v0.1.0.spkg** file can then be removed and the standard .spkg updated instead.

maoueh (changes requested):

maoueh (approved):

PR #916, opened 2026-09-01
Explain why prometheus-exporter marks an endpoint unavailable

maoueh:
An alert firing on `substreams_healthcheck_status` gave no way to tell an unreachable endpoint from an unauthenticated, overloaded or merely late one, and a flapping endpoint could produce no logs at all. This adds a failure taxonomy, separates connection setup from the request, and logs every failed poll.

- Every failure is classified into a `reason` (`invalid_config`, `connect_failed`, `connect_timeout`, `invalid_request`, `request_timeout`, `stream_error`, `stale_block`, `invalid_response`, `no_data`), exposed on a new `substreams_healthcheck_failure_count{reason,grpc_code}` counter and carried in the logs. Metrics are declared through `dmetrics`.

- Connection establishment gets its own `--connect-timeout` (default 10s), separate from `--timeout`, which now covers the `Blocks` request alone. gRPC dials lazily, so DNS/TLS/LB resolution was previously charged to the request budget and a slow connection was reported as an endpoint failure — the source of the `waiting for new LB policy update: context deadline exceeded` errors. The exporter now waits for the channel to be `READY` first, and reports `substreams_healthcheck_connect_duration_ms` and `substreams_healthcheck_stream_duration_ms` separately; `substreams_healthcheck_duration_ms` keeps its old meaning of the two combined.

- A dial that fails outright is reported as `connect_failed` within milliseconds, carrying the real dial error (`connection refused`, DNS failure), rather than waiting out the connect budget. `connect_timeout` is reserved for a connection that is merely slow to come up, and its error names the budget that expired instead of saying `context deadline exceeded`.

- Every failed poll is logged, not just the transition into `unavailable`, with reason, gRPC code, both durations and the consecutive failure count; recovery logs the downtime and how many polls failed. A block age crossing half of `--max-freshness` is reported too, so an alert on `substreams_healthcheck_block_age_ms` is no longer silent — edge-triggered and confirmed over three polls, so a chain whose block interval straddles the threshold stays quiet.

- New `substreams_healthcheck_consecutive_failures` gauge to alert on instead of `status` when single-poll hiccups should be ignored, and `block_age_ms` now resets to `NaN` when a poll returns no block instead of reporting the age of the last block ever seen.

- The exporter speaks `sf.substreams.rpc.v4.Stream/Blocks` only. The v3-to-v2 fallback is removed — it closed the connection and then kept reading from it — and `--force-protocol-version` now accepts only `4` (or unset), the flag being kept for the protocol versions to come. Nothing in production passes it today, so there is no deploy-order step.

- Fixes a panic on inconsistent label cardinality when endpoints are given different sets of query-parameter labels, and guards the block `Clock` dereference that would otherwise take the whole exporter down.

Verified end-to-end against the live fleet with a real API key: the v4 success path works on `mainnet.eth` and `mainnet.sol`, and `connect_failed` (refused port and NXDOMAIN), `connect_timeout`, `stream_error`/`Unauthenticated`, `stale_block`, the block-age crossing and the version-flag rejection were each exercised.


tools/prometheus-exporter.go
```diff
@@ -225,13 +261,13 @@ func runPrometheus(cmd *cobra.Command, args []string) error {
...
+		go launchSubstreamsPoller(endpoint, substreamsClientConfig, pkgBundle.Package, outputStreamName, startBlock, interval, connectTimeout, timeout, fresh)
 	}
 
+	// The exporter serves only its own metrics, so the collectors go to a dedicated registry
+	// instead of the global one that `dmetrics.Set.Register` would use.
 	promReg := prometheus.NewRegistry()
-	promReg.MustRegister(status)
-	promReg.MustRegister(requestDurationMs)
-	promReg.MustRegister(blockAgeMs)
+	promReg.MustRegister(initHealthcheckMetrics(labelNames)...)
```
GabrielCartier: Blocking, and a regression from the current code.

`status`, `requestDurationMs`, `blockAgeMs` used to be initialized at declaration. Now they are nil until `initHealthcheckMetrics` runs here — but the `go launchSubstreamsPoller(...)` loop above already started the pollers at line 264.

That is a data race on all seven package-level metric pointers, and when a poller wins it is a nil deref:

```
WARNING: DATA RACE
Read at 0x000108255a28 by goroutine 56:
  tools.markFailure()             tools/prometheus-exporter.go:359
  tools.launchSubstreamsPoller()  tools/prometheus-exporter.go:418
Previous write at 0x000108255a28 by main goroutine:
  tools.initHealthcheckMetrics()  tools/metrics.go:76
  tools.runPrometheus()           tools/prometheus-exporter.go:270
```

It is not theoretical: `client.NewSubstreamsClientConn` rejects an endpoint with no `:port` suffix, and rejects `--insecure --plaintext` together. Both return in microseconds, so `markFailure` reaches `status.SetInt` before this line executes. `dmetrics.GaugeVec.SetInt` is `g.p.WithLabelValues(...)` on a nil receiver, and the pollers have no `recover`, so one misconfigured endpoint kills the process and every other endpoint's metrics with it.

Hoist the `initHealthcheckMetrics` call above the endpoint loop and keep only `promReg.MustRegister(collectors...)` here.

tools/prometheus-exporter.go
```diff
@@ -247,170 +283,247 @@ func runPrometheus(cmd *cobra.Command, args []string) error {
...
+	for {
+		state := conn.GetState()
+		switch state {
+		case connectivity.Ready:
+			return nil
+		case connectivity.Shutdown:
+			return fmt.Errorf("connection shut down before becoming ready")
+		}
+
+		if !conn.WaitForStateChange(ctx, state) {
```
GabrielCartier, line 475: Blocking — this is where the PR's own goal gets lost for the most common failure.

`WaitForStateChange` returns true on any state change, and gRPC re-dials on backoff, so `TRANSIENT_FAILURE` loops here instead of terminating. A hard, instantaneous failure burns the full `--connect-timeout` and then reports `connect_timeout` with the actual dial error thrown away — `context.Cause` only yields the synthetic message.

With `--connect-timeout 4s`, a dead port and a bad hostname are indistinguishable:

```
{"endpoint": "nxdomain-abc-xyz.invalid:443", "reason": "connect_timeout", "grpc_code": "none",
 "connect_duration": "4.00128025s", "error": "connection stuck in state \"TRANSIENT_FAILURE\": connect timeout of 4s reached"}
{"endpoint": "127.0.0.1:19999", "reason": "connect_timeout", "grpc_code": "none",
 "connect_duration": "4.001306542s", "error": "connection stuck in state \"TRANSIENT_FAILURE\": connect timeout of 4s reached"}
```

On develop both come back as gRPC `Unavailable` with `transport: Error while dialing: dial tcp ...: connect: connection refused` and `... no such host`. So for "the endpoint is down" — the case this exporter exists to catch — the change is a net loss of information.

Suggest failing fast on `connectivity.TransientFailure` rather than waiting out the deadline, and surfacing the underlying dial error. `conn.WaitForStateChange` won't hand it to you directly; taking the error from a short-deadline RPC, or letting the first `TRANSIENT_FAILURE` fall through to the Blocks call, both preserve it.

Related: with this loop swallowing real connect failures, `reasonConnect` is now only reachable from `newConnection`'s synchronous config validation. `reason="connect"` reads as "could not connect" but means "CLI is misconfigured" — worth renaming, or the taxonomy misleads exactly where it is supposed to help.

tools/prometheus-exporter.go
```diff
@@ -247,170 +283,247 @@ func runPrometheus(cmd *cobra.Command, args []string) error {
...
+		result.connectDuration = time.Since(connectBegin)
+		result.reason, result.err = reasonConnectTimeout, err
+		return
+	}
+	result.connectDuration = time.Since(connectBegin)
+
+	streamBegin := time.Now()
+	defer func() { result.streamDuration = time.Since(streamBegin) }()
+
+	ctx, cancel := context.WithTimeoutCause(context.Background(), pollingTimeout, fmt.Errorf("request timeout of %s reached", pollingTimeout))
```
GabrielCartier, line 532: This cause is never read. `context.Cause` appears in exactly one place — `waitForConnReady` at line 405. On the request path gRPC returns its own `status.Error(DeadlineExceeded, "context deadline exceeded")`, which is what `classifyStreamError` and the log actually see.

So the PR body's "both deadlines carry a cause, so the error names which budget was exceeded" holds for the connect deadline only. Either read the cause when the request deadline fires, or drop the claim.

tools/prometheus-exporter.go
```diff
@@ -174,36 +196,50 @@ func runPrometheus(cmd *cobra.Command, args []string) error {
...
+	connectTimeout := sflags.MustGetDuration(cmd, "connect-timeout")
 	timeout := sflags.MustGetDuration(cmd, "timeout")
+
 	protocolVersionFlag := sflags.MustGetInt(cmd, "force-protocol-version")
 	forceProtocolVersion, err := client.ParseProtocolVersion(protocolVersionFlag)
+	if err != nil {
+		return fmt.Errorf("invalid --force-protocol-version: %w", err)
+	}
+	if !forceProtocolVersion.IsUnset() && !forceProtocolVersion.IsV4() {
+		return fmt.Errorf("invalid --force-protocol-version %d: the prometheus exporter only speaks %s for now", protocolVersionFlag, client.ProtocolVersionV4)
```
GabrielCartier: Two things.

The error message contradicts the one immediately above it: `--force-protocol-version 5` fails inside `client.ParseProtocolVersion` with `only 0, 2, 3 and 4 are supported`, then this line rejects 2 and 3. An operator reading the first message will try a value the next line refuses.

More importantly this is a breaking change with a deploy-order hazard. A deployment currently pinning `--force-protocol-version 2` or `3` goes from a working exporter to a crash loop, so the fleet loses healthcheck metrics entirely rather than degrading. It is in the changelog, but it belongs in the PR body as an explicit "unset the flag before rolling the image" step.

tools/prometheus-exporter.go
```diff
@@ -247,170 +283,247 @@ func runPrometheus(cmd *cobra.Command, args []string) error {
...
+			result.reason = reasonStaleBlock
+			result.err = fmt.Errorf("block %d is too old: %s, above the %s max freshness", clock.Number, age, *maxFreshness)
+			return
 		}
 
-		connClose()
-		cancel()
+		// A block age climbing towards the threshold is what precedes a `stale_block`
+		// failure, reporting it here is what makes an alert on `block_age_ms` explainable.
+		if age > *maxFreshness/2 {
```
GabrielCartier: This fires at `Info` on every successful poll, with no edge-trigger or rate limit. On a chain whose block interval sits near half of `--max-freshness` (default 2m), that is one line every `--interval` (20s) per endpoint, forever, for a perfectly healthy endpoint.

The failure logging elsewhere in this PR is fine — bounded by the poll interval and genuinely actionable. This one should be edge-triggered the same way the availability transitions are: log on crossing the threshold, and on recovery, not on every poll while above it.

tools/prometheus-exporter.go
```diff
@@ -247,170 +283,247 @@ func runPrometheus(cmd *cobra.Command, args []string) error {
...
-				zlog.Error("call sf.substreams.rpc.v2.Stream/Blocks", zap.String("endpoint", endpoint), zap.Error(err))
-				markFailure(endpoint, begin, err)
-				connClose()
-				cancel()
-				continue
-			}
+		// Items are ordered by block number ascending, the last one is the freshest, which is
+		// what a HEAD healthcheck cares about.
+		clock := data.BlockScopedDatas.Items[len(data.BlockScopedDatas.Items)-1].Clock
+		result.blockNum = clock.Number
```
GabrielCartier: Two small ones here.

`data.BlockScopedDatas.Items[len-1].Clock` is dereferenced unguarded — a server sending `BlockScopedData` with no `Clock` panics the poller, and with no `recover` that is the whole exporter, not one endpoint. The shape predates this PR, but given the PR is about hardening failure handling it is worth a nil check.

Also `result.blockNum` is assigned and never read anywhere (`grep -rn '\.blockNum' tools/` only finds the assignment). Same for `endpointSpecs.url` — written at line 238, never read.

tools/prometheus-exporter.go
```diff
@@ -247,170 +283,247 @@ func runPrometheus(cmd *cobra.Command, args []string) error {
...
+			fields = append(fields, zap.Duration("available_for", time.Since(state.since)))
+		}
+		zlog.Info("endpoint now marked as unavailable", fields...)
+
+		state.known = true
+		state.available = false
+		state.since = time.Now()
+	} else {
+		// Logged on every single failure, not only on the transition: an endpoint that fails
+		// repeatedly, or one that flaps between two scrapes, is otherwise invisible in the logs.
```
GabrielCartier: "not only on the transition" is diff-relative — it explains what changed rather than what the code does, and it will read as leftover commit-message text in six months. The rest of the comment carries the why on its own and is worth keeping.

Same note applies more broadly to `tools/metrics.go`: the seven `failureReason` constants each carry a doc comment where the identifier plus the changelog already says it.

GabrielCartier (changes requested):
Solid direction and the failure taxonomy is the right shape — the cardinality fix is genuinely correct, and I verified it at runtime with endpoints carrying different query-param label sets. Two things need to change before this goes in.

The first is a crasher: the metric pointers are now assigned by `initHealthcheckMetrics`, but the poller goroutines are launched before that call. Any endpoint whose first poll fails synchronously inside `NewSubstreamsClientConn` (missing `:port`, or `--insecure --plaintext` together) reaches `markFailure` while `status` is still nil. `dmetrics.GaugeVec.SetInt` derefs immediately and the pollers have no `recover`, so it takes the whole exporter down, all endpoints. `go build -race` flags it too. Reproduced both halves.

The second is that the change loses the "why" for the most common failure. `waitForConnReady` treats `TRANSIENT_FAILURE` as retryable, so connection-refused and NXDOMAIN both burn the full `--connect-timeout` and come out as byte-identical `connect_timeout` lines with the real dial error discarded. Before this, both surfaced as gRPC `Unavailable` carrying `dial tcp ...: connection refused` / `no such host`. Endpoint-down is the case an operator hits most, and it's the one that got harder to diagnose.

Rest is smaller — details inline. Worth calling the `--force-protocol-version` rejection out in the PR body with a deploy-order note: anything currently pinning 2 or 3 crash-loops on the new image, which silences that fleet's healthcheck metrics at exactly the wrong moment.

tools/prometheus-exporter.go
maoueh: Applied, with one correction from testing it: a bare edge trigger still logs once per poll on exactly the chain you described. Against `mainnet.eth` with `--max-freshness 40s` the age oscillated 14s–26s across the 20s threshold, so it alternated climbed/fell every poll — same noise, in pairs. Added a 3-consecutive-poll confirmation streak; that case now logs nothing, and a sustained drift logs once.

tools/prometheus-exporter.go
maoueh: Applied both. Fails fast on `TRANSIENT_FAILURE` and lets the request surface the dial error; a refused port now reports `connect_failed` / `Unavailable` in ~2ms with `dial tcp 127.0.0.1:19999: connect: connection refused` instead of burning the full budget. One caveat: NXDOMAIN comes back as `no children to pick from` rather than `no such host` — the roundrobin balancer replaces the resolver error. Still fast and distinguishable, but not quite develop’s message. `reasonConnect` is renamed `invalid_config`.

tools/prometheus-exporter.go
maoueh: Walking part of this back: the flag is not passed by any production deployment, confirmed by @maoueh, so there is no deploy-order step to call out. Removed the warning from the PR body and trimmed the changelog to the plain breaking note.

tools/prometheus-exporter.go
```diff
@@ -247,170 +293,309 @@ func runPrometheus(cmd *cobra.Command, args []string) error {
...
+// backoff, so waiting through TRANSIENT_FAILURE would burn the whole connect budget on an
+// endpoint that answered "connection refused" in a microsecond, and report it as a timeout.
+func waitForConnReady(ctx context.Context, conn *grpc.ClientConn) error {
+	conn.Connect()
+	for {
+		state := conn.GetState()
+		switch state {
+		case connectivity.Ready:
+			return nil
+		case connectivity.TransientFailure:
```
GabrielCartier: Blocking — this fixes the diagnostic but regresses the availability signal.

`TRANSIENT_FAILURE` is not terminal. gRPC enters it on any failed dial attempt and re-dials on its own backoff, so returning here gives up on connections that were about to succeed. I ran the develop loop and this one against a real `grpc.Server` that starts listening at t=2s, with a 10s budget:

```
OLD (server up at t=2s, budget 10s)  elapsed=2.86s  err=<nil>              <- poll SUCCEEDS
NEW (server up at t=2s, budget 10s)  elapsed=0.00s  err=connection failed to establish
```

So any rolling restart of a Substreams backend, any DNS blip, any LB moment with zero healthy backends now flips `substreams_healthcheck_status` to 0 and increments `failure_count{reason="connect_failed"}` on the first refused SYN. `--connect-timeout` has become dead budget for the entire class it was introduced to cover, and the flag's own help text at line 60 still promises the opposite — "endpoints will be considered 'failing' ... if the gRPC connection does not become ready in that duration".

What I think you want is to keep waiting through `TRANSIENT_FAILURE` — that is what the budget is for — while remembering that a dial actually failed, so the caller still issues the request and gets the real error instead of a bare `connect_timeout`:

```go
func waitForConnReady(ctx context.Context, conn *grpc.ClientConn) error {
	conn.Connect()

	dialFailed := false
	for {
		state := conn.GetState()
		switch state {
		case connectivity.Ready:
			return nil
		case connectivity.TransientFailure:
			// gRPC re-dials on its own backoff, so this is not terminal: an endpoint restarting
			// behind a load balancer refuses the first dial and accepts the next one. Waiting is
			// what the connect budget is for, but the dial error is worth keeping.
			dialFailed = true
		case connectivity.Shutdown:
			return fmt.Errorf("connection shut down before becoming ready")
		}

		if !conn.WaitForStateChange(ctx, state) {
			if dialFailed {
				return errConnFailedFast
			}
			return fmt.Errorf("connection stuck in state %q: %w", state, context.Cause(ctx))
		}
	}
}
```

You keep `connect_failed` with the real dial error, and `connect_timeout` now means what it says: never failed, just never got there. Both regression tests pass with this, and the full `tools` suite stays green under `-race`.

The trade-off is explicit and yours to make: a permanently dead endpoint costs the full `--connect-timeout` again rather than 2ms. For a healthcheck polling every 20s I would take that over false unavailability during restarts, but if the 2ms matters, a bounded grace (wait through the first N failures, or a fraction of the budget) gets most of both. What should not stand is the current behaviour, where the budget is never spent at all.

docs/release-notes/change-log.md
```diff
@@ -98,6 +98,48 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 - `substreams-tier1` now names the usage marker it writes in every module cache folder after the request's plan tier: `last_used_<plan>` (lowercase, e.g. `last_used_pro`), still plain `last_used` when unauthenticated. `firecore tools substreams purge` reads the plan back from that name to apply a retention per plan.
 
+### Tools
+
+- `substreams tools prometheus-exporter` now says *why* an endpoint is down. Every failure is classified into a
+  `reason` -- `invalid_config`, `connect_failed`, `connect_timeout`, `invalid_request`, `request_timeout`,
+  `stream_error`, `stale_block`, `invalid_response` or `no_data` -- exposed on the new
+  `substreams_healthcheck_failure_count{reason,grpc_code}` counter and included in the logs. An alert firing on
+  `substreams_healthcheck_status` no longer requires guessing whether the endpoint was unreachable, unauthenticated,
+  overloaded or merely late. A dial that fails outright is reported as `connect_failed` within milliseconds, carrying
```
GabrielCartier: One open question here, now that the DNS claim is corrected (`1118bbbb`).

"`connect_timeout` is reserved for a connection that is merely slow to come up" reads as a naming refinement, but as currently implemented it is a behaviour change: `connect_timeout` is nearly unreachable, because anything that fails a dial short-circuits before the budget is ever spent. Same for the next bullet's "The exporter now waits for the channel to be `READY` before issuing the request" — it doesn't, when the dial fails.

If you take the `waitForConnReady` change, both sentences become true as written and nothing more is needed. If you keep fail-fast, the availability-semantics change wants its own bullet — it is the part an operator would want to read before upgrading.

tools/prometheus-exporter.go
```diff
@@ -247,170 +293,309 @@ func runPrometheus(cmd *cobra.Command, args []string) error {
...
+		return
+	}
+
+	aboveHalf := *result.blockAge > *result.maxFreshness/2
+	if aboveHalf == state.blockAgeAboveHalf {
+		state.blockAgeStreak = 0
+		return
+	}
+
+	state.blockAgeStreak++
```
GabrielCartier: `blockAgeStreak` is not reset when `blockAge` is nil, so the early return at line 399 lets the streak survive a gap.

Sequence: two polls above half, then five failed polls returning no block data, then one more above half — the counter hits 3 and the log says `confirmed_over_polls: 3`, describing three consecutive agreeing polls across a seven-poll window with an outage in the middle.

The streak exists precisely to mean "this is sustained, not oscillation", so a reset alongside the nil check would keep it honest.

tools/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// streamFailure attributes an error returned by the Blocks call. `connect_failed` means the
+// dial itself failed, `connect_timeout` (set by the caller) means it was merely slow and the
+// backend was never reached at all. When the channel never
+// became ready, gRPC answers with the dial error it was holding, so the failure belongs to
+// the connection rather than to the endpoint's own answer.
+func streamFailure(ctx context.Context, connectFailedFast bool, err error) (failureReason, error) {
+	err = withDeadlineCause(ctx, err)
+	if connectFailedFast {
```
GabrielCartier: `connectFailedFast` short-circuits before the error is looked at, so the reason and the code can disagree.

Once it is set at `prometheus-exporter.go:519`, gRPC keeps re-dialing on its backoff. If the channel reaches READY before the RPC is issued and the stream then fails for an unrelated reason — `ResourceExhausted`, `Unauthenticated`, a request deadline — this still returns `reasonConnectFailed` while `grpcCodeOf` reports the true code. `failure_count{reason="connect_failed",grpc_code="Unauthenticated"}` is a series that describes nothing real.

Low probability today (gRPC's base backoff is ~1s against microseconds to issue the RPC), and it shrinks further if you take the wait-through-`TRANSIENT_FAILURE` version, since the flag would then only be set at budget expiry. Left unpatched for that reason — worth settling `waitForConnReady` first, then gating this on the error rather than the flag alone.

tools/prometheus-exporter.go
```diff
@@ -174,36 +198,58 @@ func runPrometheus(cmd *cobra.Command, args []string) error {
...
+
+	// Checked before `ParseProtocolVersion` so that an operator passing 2 or 3 reads the one
+	// message that applies here, rather than being told v2 and v3 are supported and then
+	// refused them on the next line.
 	protocolVersionFlag := sflags.MustGetInt(cmd, "force-protocol-version")
+	if protocolVersionFlag != 0 && protocolVersionFlag != int(client.ProtocolVersionV4) {
+		return fmt.Errorf("invalid --force-protocol-version %d: the prometheus exporter only speaks %s for now, leave the flag unset or pass 4", protocolVersionFlag, client.ProtocolVersionV4)
+	}
+
 	forceProtocolVersion, err := client.ParseProtocolVersion(protocolVersionFlag)
```
GabrielCartier: `subReq.Validate()` runs inside the poll loop, after a full dial, on every poll of every endpoint.

A bad module name is a startup config error. Hoisting the validation next to the other startup checks in `runPrometheus` would fail the command immediately, instead of reporting every endpoint as `invalid_request` forever, once per interval, each time paying a TCP+TLS connection first.

Left this one to you — it changes when the command fails, which is more than a nit.

GabrielCartier (changes requested):
Second pass. Six of the seven are properly resolved and I verified each rather than taking the replies at face value — init ordering exercised under `-race` with both trigger cases, `context.Cause` confirmed end-to-end (`dgrpc.AsGRPCError` unwraps, so the `grpc_code` label survives the `%w`), the freshness streak traced through sustained drift / straddling / recovery / nil `maxFreshness`, and the nil-`Clock` branch confirmed reachable against a fake server. All nine taxonomy reasons are reachable, none dead. CI green including `-race`.

The `waitForConnReady` fix is the problem. Failing fast on the first `TRANSIENT_FAILURE` fixed the diagnostic but broke recovery: gRPC enters that state on any failed dial and re-dials on its own backoff, so it is not terminal. An endpoint that refuses one dial and accepts the next — a rolling restart, a brief DNS blip, a load balancer moment with no healthy backend — is now reported unavailable on the first refused SYN, and `--connect-timeout` no longer covers the case it exists for. That is a behaviour change to the availability signal itself, not to its explanation, which makes it worse than the bug it replaced.

I wrote two regression tests rather than just asserting this. Both fail on `06d6569b`:

```
--- FAIL: TestWaitForConnReady_RecoversWithinConnectBudget (0.50s)
    endpoint came up after 500ms, well inside the 5s connect budget,
    but the poll gave up after 593.833µs
--- FAIL: TestWaitForConnReady_GivesUpAfterConnectBudget (0.00s)
    gave up after 529.125µs without spending the 500ms connect budget
```

Both pass with the fix suggested inline, and the full `tools` suite stays green under `-race`. Test file and the one-function change are in the inline comments — take them or replace them, but `waitForConnReady` should not stay uncovered either way: it is the site of both this fix and the previous one, and neither had a test that could fail.

The rest is small: the changelog claims the DNS failure is carried in the error and it isn't, and a few nits.

One aside — `44c481de`'s message body ends with the `git merge` comment template committed verbatim (`# Conflicts:` / `#	docs/release-notes/change-log.md`). Worth a reword on the next force-push.

tools/prometheus-exporter.go
GabrielCartier: Pushed the two regression tests for this as `769d70a0` (`tools/prometheus_exporter_connready_test.go`) rather than leaving them in a comment — they fail on the current head, so CI is red until the behaviour changes, which is the point of them.

`TestWaitForConnReady_RecoversWithinConnectBudget` is the regression: backend starts listening 500ms in, 5s budget, poll gives up in ~600µs. `TestWaitForConnReady_GivesUpAfterConnectBudget` pins the other half — a backend down longer than the budget must still be reported, but the budget has to be *spent*, not skipped. That second one is what makes the difference between fail-fast and wait observable in CI instead of in production.

I did not push the `waitForConnReady` change itself. The trade-off is yours: waiting through `TRANSIENT_FAILURE` costs a dead endpoint the full connect budget again instead of 2ms, and a bounded grace (first N failures, or a fraction of the budget) gets most of both. The suggested implementation in my review comment above turns both tests green with the full `tools` suite passing under `-race`, if you want it as-is.

tools/prometheus-exporter.go
GabrielCartier: Follow-up on the streak: pushed `6686ef61`.

`trackBlockAge` returned early when `blockAge` was nil without clearing `blockAgeStreak`, so a streak could span an outage — two polls above half, five failed polls carrying no age, then one more above half, and the log claimed `confirmed_over_polls: 3` for a window that was never consecutive. The reset is in the nil branch now.

The streak logic itself is right, and the 3-poll confirmation is a better answer than the bare edge trigger I suggested — the oscillation you measured on `mainnet.eth` is exactly the case an edge trigger gets wrong.

tools/prometheus-exporter.go
GabrielCartier: The contradictory-message half is fixed, and the walk-back on the deploy note is fair enough given nothing passes the flag.

One leftover, pushed as `dae4ae6c`: with the new pre-check rejecting everything but 0 and 4, and both of those parsing cleanly, the `if err != nil` on `ParseProtocolVersion` below it could no longer be reached. Dropped it and left a line saying why the error is discarded.

GabrielCartier:
Pushed four commits rather than leaving the code in comments:

- `769d70a0` — `test(tools)`: the two `waitForConnReady` regression tests. **These fail on the current head**, so CI is red until the connect-budget behaviour changes. That is deliberate: the red is the finding.
- `6686ef61` — `fix(tools)`: reset the block-age streak on polls carrying no age, so a streak cannot span an outage.
- `dae4ae6c` — `refactor(tools)`: drop the now-unreachable `ParseProtocolVersion` error branch.
- `1118bbbb` — `docs(tools)`: correct the changelog's DNS claim and the garbled `streamFailure` doc comment.

Each is one logical change; `go build`, `go vet` and `gofmt` are clean, and the only failing tests are the two new ones.

What I did **not** push, on purpose: the `waitForConnReady` fix itself, the `connectFailedFast` classification gate, and hoisting `subReq.Validate()` out of the poll loop. The first is a genuine trade-off (a dead endpoint costs the full connect budget again instead of 2ms) and the other two hang off however you resolve it. Suggested implementation is in the thread if you want it as-is — it turns both tests green with the suite passing under `-race`.

I've cleared out the review comments that the pushes resolved, so what's left inline is only what's still open. I also withdrew an earlier nit about the `test-go/testify` → `stretchr/testify` swap: `extract-proto_test.go` and `metrics_test.go` already use `stretchr`, so the change makes the file match its package. That one was my mistake.

Happy to take the `waitForConnReady` change too if you'd rather I just applied it — say which shape you want.

tools/prometheus-exporter.go
GabrielCartier: Verified `ebbe7294` locally, not just from the CI summary — both regression tests pass, full `tools` suite green (35) under `-race`, build/vet/gofmt clean:

```
--- PASS: TestWaitForConnReady_RecoversWithinConnectBudget (1.00s)
--- PASS: TestWaitForConnReady_GivesUpAfterConnectBudget (0.50s)
```

The shape is right: `dialFailed` is remembered but the loop still waits out the budget, so a backend that comes back inside it is reported healthy, while one that failed a dial still gets `connect_failed` with the real error rather than a bare timeout. `connect_timeout` keeps its meaning for a channel that never left `CONNECTING`. `errConnDialFailed` is a better name than mine — "fast" was the part that was wrong.

Resolving.

docs/release-notes/change-log.md
GabrielCartier: Resolved by `ebbe7294`. The sentence I left in place — "`connect_timeout` is reserved for a connection that is merely slow to come up and never failed a dial" — is now accurate as written, and "waits for the channel to be `READY` before issuing the request" is true again too. Nothing further needed here.

GabrielCartier (approved):
Approving. Verified `ebbe7294` locally rather than off the CI summary: both regression tests pass, full `tools` suite green (35) under `-race`, build/vet/gofmt clean.

The connect-budget fix is the right shape — `dialFailed` is remembered while the loop still waits out the budget, so a backend that recovers inside it reports healthy and one that genuinely failed a dial still gets `connect_failed` with the real error instead of a bare timeout. `connect_timeout` keeps its meaning for a channel that never left `CONNECTING`. `errConnDialFailed` is a better name than the one I suggested; "fast" was the part that was wrong.

I'm leaving the two remaining threads open on purpose — `streamFailure` short-circuiting on `connectFailed` without consulting the error, and `subReq.Validate()` running per-poll after a full dial. Neither is necessary for this to merge. Take them or close them as you like.

PR #921, opened 2026-09-02
Count requests that are still setting up in the eviction autoscaler metric

sduchesneau:
`substreams_tier1_effective_active_requests` could read *below* `substreams_active_requests`, the metric it is meant to replace as the horizontal autoscaler input. It is a `max()` of the active-request count and the CPU-derived count, so that should not be possible.

The two were counting different sets of requests. `substreams_active_requests` increments early in `blocks()`, while a request only enters the evictor's manager map after session acquisition, the deterministic-error check against the cache store and store cloning. Every request still in setup was counted by one and invisible to the other. Teardown has the same shape: the deferred `manager.Remove` is registered later than the deferred `Tier1ActiveRequests.Dec()`, so it runs first.

The evictor now takes its active-request count from the same accessor admission uses (`getActiveRequestCount`), through a `CountActiveRequestsWith` hook, falling back to the manager map when unset. The manager map is still the source of eviction candidates — a request in setup has no stats to sample and nothing worth cancelling.

Effect: a pod with requests queued in setup no longer looks emptier to the autoscaler than it is, and the eviction metric agrees with the soft and hard admission limits about how many requests exist.


UlysseCorbeil (approved):
lgtm

PR #922, opened 2026-09-03
Overlap store reads and sends when tier1 streams cached execution output

sduchesneau:
Four tier1 changes to the production-mode streaming path.

**Prefetch cached execution output files.** Each 1000-block segment was opened, decompressed and sent before the next one was requested from the object store, so every segment paid a full store round trip on the critical path. Tier1 now downloads the next segments while streaming the current one, bounded per request by `Tier1Config.ExecOutPrefetch`: at most 4 segments ahead holding at most 64 MiB of decompressed data, and that budget is a hard cap. No size is asked of the store: the decompressed size of the last downloaded segment is the estimate for the next ones, as many download at once as estimate-sized files fit in the budget, and each in-flight download reserves an even share of the budget as its read limit, so in-flight reads never add up to more than the budget. A file over its share is left to the walker and doubles the estimate, so a chain going from quiet to busy shrinks the concurrency instead of overshooting. A file over the whole budget turns prefetching off for that request. Missing files are left to the walker's existing retry loop. Off in noop mode.

**Send batches on their own goroutine.** The walker built a batch, sent it, then decoded the next one, so gRPC compression and the socket write sat on the critical path of every batch. Sends now go through an unbuffered channel to a sender goroutine, so one batch is compressed and written while the next is decoded. A failed send is sticky and nothing goes out after it. A panic inside a send is recovered into a send error, as the walker's own goroutine did before. The walker drains the sender before a segment is reported done, so the linear pipeline can never send ahead of cached data and message order is unchanged. Measured on a synthetic 10 MiB segment: decode 1.3 ms, marshal plus zstd 5.9 ms, serial 7.2 ms, overlapped 6.5 ms. The send side dominates, and most of it is zstd.

**Stop copying payloads while decoding.** `ReadNextItem` reads each item into a fresh buffer nothing else touches, so the item now aliases it instead of copying the payload out.

**Write cache markers off the critical path.** The `substreams.spkg` and `last_used` writes ran synchronously before the pipeline started. They now run fire and forget on a detached context with a 30 s timeout. The request neither starts nor exits waiting for them, so a hanging store cannot hold a cancelled request's slot. The integration tests poll for the spkg marker instead of expecting it the moment the request returns.

Ordering is preserved at every step: the walker still asks for one segment at a time and only advances on a completed send, the prefetcher only answers for the segment asked, and the segmenter is immutable. Files are created atomically and never overwritten, so reading them earlier cannot pick up different content.

Tests: prefetch tests cover ordering, depth and budget bounds, the share overflow path with a held-bytes probe that never exceeds the budget, overflow of the whole budget, a missing segment and cancellation. The walker had no tests before; new ones cover block order under a slow sink in buffered and unbuffered modes, a send error stopping the segment, a panic in a send, and the exclusive end block. All pass under the race detector, 20 repetitions for the prefetch suite.

One measured non-change: egress accounting calls `proto.Size` before `SendMsg`, which looked like a double marshal. On a 100-block, 10 MiB batch it costs 11 µs against 812 µs for the marshal, so it was left alone.

### DIFF
--- a/app/tier1.go
+++ b/app/tier1.go
@@ -27,6 +27,7 @@ import (
 	"github.com/streamingfast/substreams/reqctx"
 	"github.com/streamingfast/substreams/service"
 	"github.com/streamingfast/substreams/service/active_requests"
+	"github.com/streamingfast/substreams/storage/execout"
 	"github.com/streamingfast/substreams/wasm"
 	_ "github.com/streamingfast/substreams/wasm/wasmtime"
 	"github.com/streamingfast/substreams/wasm/wazero"
@@ -62,6 +63,7 @@ func NewDefaultTier1Config() *Tier1Config {
 		MergedBlocksBundleSize: bstream.DefaultMergedBlocksBundleSize,
 		BlockExecutionTimeout:  1 * time.Minute,
 		OutputBufferSize:       100,
+		ExecOutPrefetch:        execout.PrefetchConfig{Depth: execout.MaxPrefetchDepth, BudgetBytes: 64 << 20},
 	}
 }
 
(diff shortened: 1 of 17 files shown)


sduchesneau (commented):
Gives me good +20% improvements for precached stuff

UlysseCorbeil (approved):

PR #924, opened 2026-09-04
Webhook sink: auth, signature, resumable delivery, reorg notifications, batching

sduchesneau:
Five independent commits to `substreams sink webhook`, in review order:

1. **Auth header and HMAC signature.** `--webhook-auth-header-name` / `--webhook-auth-header-value-envvar` send a verbatim header; `--webhook-signing-secret-envvar` adds `X-Substreams-Signature: t=<unix>,v1=<hex>` over `<t>.<body>`. Secrets come from env vars so they never sit on the command line. `Client.Call` now returns a `*DeliveryError` with status and attempt count once retries run out.
2. **Resumable delivery.** `--webhook-on-failure=exit` writes the failed payload to `<state-file>.pending`, writes a JSON reason to `--webhook-termination-log` when that file exists, and exits with status 75. On the next start the pending payload is delivered before any Substreams stream is opened, so retrying against a dead endpoint costs no egress. The file is only written on failure: a kill mid-call leaves nothing behind, the cursor was not advanced, and the stream re-sends that block. A changed URL or secret resets `first_attempt_at`. Also exposes `substreams_sink_progress_block`.
3. **Sink library fix.** With `--undo-buffer-size`, an undo signal reaching below the buffer while nothing had been emitted yet was swallowed. That is what a restart from a cursor on a fork produces, so the previous run's forked blocks were never rolled back. The buffer now reports whether it absorbed the undo and the sinker forwards it to the handler otherwise. Affects every buffered sink, not just the webhook one.
4. **Reorg notifications.** `--webhook-undo-url` receives `{"lastValidBlock": {"number", "id"}, "manifest": {"moduleName"}}` per undo signal, with the same headers, retry, on-failure and pending-file rules as blocks. Opt-in; without it the cursor moves back and only replacement blocks are delivered, as before.
5. **Batching.** `--webhook-batch-max-blocks=N` switches every call to `{"manifest": {...}, "blocks": [{"clock", "data"}, ...]}` with up to N blocks. A batch is sent when full, when `--webhook-batch-max-wait` (1s) has passed and the next block arrives, when the chain is live, before an undo notification, and when the stream ends. A failed batch is kept and resumed as one payload. Toggling batching while the sink is stopped discards a pending payload of the other shape and lets the stream re-send its blocks. Off by default.

Not included: a receive-window cap on the Substreams client.

Design context: streamingfast/services-control-plane#61.

### DIFF
--- a/cmd/substreams/sink_webhook.go
+++ b/cmd/substreams/sink_webhook.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"context"
+	"fmt"
+	"os"
 	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/streamingfast/cli"
 	"github.com/streamingfast/cli/sflags"
 	"github.com/streamingfast/derr"
 	"github.com/streamingfast/substreams/sink"
@@ -28,6 +31,9 @@ func init() {
 	sinkWebhookCmd.Flags().Int("webhook-max-retries", 3, "Maximum number of retries for webhook calls (0 disables retries, -1 for infinite retries)")
 	sinkWebhookCmd.Flags().Duration("webhook-timeout", 30*time.Second, "Timeout for individual webhook calls")
 	sinkWebhookCmd.Flags().Duration("webhook-max-retry-interval", 30*time.Second, "Maximum interval between webhook retries (exponential backoff cap)")
+	sinkWebhookCmd.Flags().String("webhook-auth-header-name", webhook.DefaultAuthHeaderName, "Name of the header carrying the value read from --webhook-auth-header-value-envvar")
+	sinkWebhookCmd.Flags().String("webhook-auth-header-value-envvar", "WEBHOOK_AUTH_HEADER_VALUE", "Environment variable holding the auth header value sent on every call, for example 'Bearer <token>'. No header is sent when the variable is empty")
+	sinkWebhookCmd.Flags().String("webhook-signing-secret-envvar", "WEBHOOK_SIGNING_SECRET", fmt.Sprintf("Environment variable holding the secret used to sign every call body with HMAC-SHA256 in the %s header. No signature is sent when the variable is empty", webhook.SignatureHeader))
 
 	SinkCmd.AddCommand(sinkWebhookCmd)
 }
@@ -36,8 +42,23 @@ func init() {
 var sinkWebhookCmd = &cobra.Command{
 	Use:   "webhook <url> [<manifest> [<module_name>]]",
 	Short: "Trigger a webhook call for each event from a substreams module",
-	RunE:  sinkWebhookE,
-	Args:  cobra.RangeArgs(1, 3),
+	Long: cli.Dedent(`
+		POST the output of a substreams module to <url>, one call per block, as JSON:
+
+		  {"clock": {"number": ..., "id": "...", "timestamp": "..."},
+		   "manifest": {"moduleName": "...", "type": "..."},
+		   "data": {...}}
+
+		Calls carry the header named by --webhook-auth-header-name when the variable named by
+		--webhook-auth-header-value-envvar is set, and an ` + webhook.SignatureHeader + ` header when the
+		variable named by --webhook-signing-secret-envvar is set. The signature value is
+		"t=<unix seconds>,v1=<hex>" where <hex> is HMAC-SHA256 over "<t>.<body>" keyed with the secret.
+
+		Network errors and 5xx responses are retried with exponential backoff up to --webhook-max-retries.
+		A 4xx response is not retried.
+	`),
+	RunE: sinkWebhookE,
+	Args: cobra.RangeArgs(1, 3),
 }
 
 func sinkWebhookE(cmd *cobra.Command, args []string) error {
@@ -60,21 +81,17 @@ func sinkWebhookE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// Get webhook configuration from flags
-	stateFileStr := sflags.MustGetString(cmd, "state-file")
-	webhookTimeout := sflags.MustGetDuration(cmd, "webhook-timeout")
-	webhookMaxRetries := sflags.MustGetInt(cmd, "webhook-max-retries")
-	webhookMaxRetryInterval := sflags.MustGetDuration(cmd, "webhook-max-retry-interval")
-
-	// Create webhook sink configuration
 	sinkConfig := webhook.SinkConfig{
 		WebhookURL:   webhookURL,
-		StateFile:    stateFileStr,
+		StateFile:    sflags.MustGetString(cmd, "state-file"),
 		SinkerConfig: sinkerConfig,
 		ClientConfig: webhook.Config{
-			Timeout:     webhookTimeout,
-			MaxRetries:  webhookMaxRetries,
-			MaxInterval: webhookMaxRetryInterval,
+			Timeout:         sflags.MustGetDuration(cmd, "webhook-timeout"),
+			MaxRetries:      sflags.MustGetInt(cmd, "webhook-max-retries"),
+			MaxInterval:     sflags.MustGetDuration(cmd, "webhook-max-retry-interval"),
+			AuthHeaderName:  sflags.MustGetString(cmd, "webhook-auth-header-name"),
+			AuthHeaderValue: os.Getenv(sflags.MustGetString(cmd, "webhook-auth-header-value-envvar")),
+			SigningSecret:   os.Getenv(sflags.MustGetString(cmd, "webhook-signing-secret-envvar")),
 		},
 		Logger: zlog,
 	}
--- a/cmd/substreams/sink_webhook.go
+++ b/cmd/substreams/sink_webhook.go
@@ -34,6 +34,8 @@ func init() {
 	sinkWebhookCmd.Flags().Duration("webhook-timeout", 30*time.Second, "Timeout for individual webhook calls")
 	sinkWebhookCmd.Flags().Duration("webhook-max-retry-interval", 30*time.Second, "Maximum interval between webhook retries (exponential backoff cap)")
 	sinkWebhookCmd.Flags().String("webhook-on-failure", string(webhook.OnFailureSkip), fmt.Sprintf("What to do once every retry for a block has failed: %q drops the block and continues, %q keeps the block on disk, writes the reason to the termination log and exits with status %d; the next start delivers that block before it connects to Substreams", webhook.OnFailureSkip, webhook.OnFailureExit, webhook.ExitCodeDeliveryFailed))
+	sinkWebhookCmd.Flags().Int("webhook-batch-max-blocks", 0, "Send up to this many blocks per call, in the batch payload shape (see below). 0 sends one block per call in the single-block shape")
+	sinkWebhookCmd.Flags().Duration("webhook-batch-max-wait", time.Second, "Longest a batch waits for more blocks before it is sent, checked when the next block arrives. A batch is also sent when the chain is live, before an undo notification, and when the stream ends")
 	sinkWebhookCmd.Flags().String("webhook-undo-url", "", "URL that receives a POST for each chain reorganization, with body {\"lastValidBlock\": {\"number\": ..., \"id\": \"...\"}, \"manifest\": {\"moduleName\": \"...\"}}. Empty disables the notification; the blocks that replace the undone ones are still delivered to <url>")
 	sinkWebhookCmd.Flags().String("webhook-termination-log", "/dev/termination-log", "File that receives the reason for a delivery-failure exit, written only when the file already exists (Kubernetes creates it)")
 	sinkWebhookCmd.Flags().String("webhook-auth-header-name", webhook.DefaultAuthHeaderName, "Name of the header carrying the value read from --webhook-auth-header-value-envvar")
@@ -46,13 +48,21 @@ func init() {
 // sinkWebhookCmd represents the command to run substreams webhook sink
 var sinkWebhookCmd = &cobra.Command{
 	Use:   "webhook <url> [<manifest> [<module_name>]]",
-	Short: "Trigger a webhook call for each event from a substreams module",
+	Short: "POST the output of a substreams module to a webhook, one call per block or per batch of blocks",
 	Long: cli.Dedent(`
 		POST the output of a substreams module to <url>, one call per block, as JSON:
 
 		  {"clock": {"number": ..., "id": "...", "timestamp": "..."},
 		   "manifest": {"moduleName": "...", "type": "..."},
 		   "data": {...}}
+
+		With --webhook-batch-max-blocks=N every call carries up to N blocks, a batch of one included, as:
+
+		  {"manifest": {"moduleName": "...", "type": "..."},
+		   "blocks": [{"clock": {...}, "data": {...}}, ...]}
+
+		Blocks are in ascending order. Switching batching on or off while the sink is stopped discards a
+		pending payload of the other shape; its blocks come back through the stream in the new shape.
 
 		Calls carry the header named by --webhook-auth-header-name when the variable named by
 		--webhook-auth-header-value-envvar is set, and an ` + webhook.SignatureHeader + ` header when the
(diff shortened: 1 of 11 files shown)


sink/webhook/sink.go
```diff
@@ -82,19 +206,173 @@ func (s *Sink) Run(ctx context.Context) error {
...
+	return cursor, nil
+}
+
+// deliver POSTs the pending payload and, on success, commits it: cursor
+// written, pending file removed, progress metric updated. The returned error
+// is the *DeliveryError from the client.
+func (s *Sink) deliver(ctx context.Context, pending *pendingDelivery) error {
+	url := s.webhookURL
+	if pending.isUndo() {
+		url = s.undoURL
```
sduchesneau: **A saved reorg notification can make the sink exit on every start.** In exit mode, a reorg notification that fails to send is written to the pending file. If the operator then restarts without `--webhook-undo-url`, `s.undoURL` is empty and the sink POSTs the saved payload to `""`. That error is retried like a network error and ends in another exit 75, on every start, until someone deletes the pending file by hand.

Suggestion: when loading a pending undo payload with no undo URL configured, either drop it with a warning or fail fast with a clear config error, instead of trying the empty URL.

---
_Generated by [Claude Code](https://claude.ai/code)_

sink/webhook/sink.go
```diff
@@ -82,19 +206,173 @@ func (s *Sink) Run(ctx context.Context) error {
...
+	if err := s.client.Call(ctx, url, pending.Payload, pending.BlockNumber); err != nil {
+		return err
+	}
+
+	if s.stateFile != "" && pending.Cursor != "" {
+		cursor, err := sink.NewCursor(pending.Cursor)
+		if err != nil {
+			return fmt.Errorf("invalid cursor for block %d: %w", pending.BlockNumber, err)
+		}
+		if err := sink.WriteCursor(s.stateFile, cursor); err != nil {
```
sduchesneau: **Failing to save the cursor after a successful POST counts as a failed delivery.** Before this PR, a cursor write failure was only a warning. Now it is returned from `deliver`, and `send` treats it as a delivery failure:
- In exit mode, the payload the receiver already accepted is saved as pending, the sink exits 75, and it sends the payload again on restart.
- In skip mode, it logs "dropping block after failed delivery" for a block the receiver actually got.
- For a reorg notification, the error reports the main webhook URL rather than the undo URL.

Suggestion: split post-POST commit errors from delivery errors, for example by returning a distinct error type, so `send` does not apply the on-failure policy to them.

---
_Generated by [Claude Code](https://claude.ai/code)_

sink/webhook/sink.go
```diff
@@ -104,43 +382,125 @@ func (s *Sink) handleBlockScopedData(ctx context.Context, data *pbsubstreamsrpc.
...
 
-	// Save cursor to state file
-	if s.stateFile != "" && cursor != nil {
-		if err := sink.WriteCursor(s.stateFile, cursor); err != nil {
-			s.logger.Warn("failed to save cursor to state file",
-				zap.Error(err),
-				zap.String("file", s.stateFile))
-		}
+	if s.onFailure == OnFailureExit {
+		return s.deliveryFailed(pending, err)
```
sduchesneau: **A shutdown during retries is handled as a delivery failure.** If the context is canceled (SIGTERM) while `client.Call` is retrying, the error lands here like any other delivery failure:
- In exit mode, it writes a pending file and a `webhook_delivery_failed` termination message, even though the endpoint never finished failing. This contradicts the help text, which says "a kill in the middle of a call leaves no pending file".
- If the SIGTERM arrives while the sink is resending a saved payload at startup, the process exits 75 instead of shutting down cleanly.
- If it arrives during normal streaming, the process exits 0, but the termination log still reports a delivery failure.
- In skip mode, the same shutdown deletes the saved payload instead of keeping it.

Suggestion: check `ctx.Err() != nil` (or `errors.Is(err, context.Canceled)`) before applying the on-failure policy, and return the context error without touching the pending file.

---
_Generated by [Claude Code](https://claude.ai/code)_

sduchesneau (commented):
I reviewed the webhook sink changes: batching, a separate URL for reorg notifications, the pending file for failed deliveries, auth and signing headers, and forwarding reorgs the undo buffer can't absorb. I found three issues, all in `sink/webhook/sink.go`, and left them as inline comments. Each one is about which errors trigger the on-failure policy:

1. A saved reorg notification loops forever with exit 75 if the sink restarts without `--webhook-undo-url`.
2. A shutdown (SIGTERM) during retries is treated as a delivery failure, so it writes a pending file or drops one.
3. A cursor write failure after a successful POST is treated as a delivery failure, so a payload the receiver already accepted gets sent again or logged as dropped.

The rest of the changed files looked fine.

---
_Generated by [Claude Code](https://claude.ai/code)_

sduchesneau:
I fixed all three review comments in `sink/webhook/sink.go` and added a regression test for each. The `sink/webhook` tests pass, and all four new tests fail against the old `sink.go`. Nothing is committed yet.

1. **Saved reorg notification with no `--webhook-undo-url`:** at startup, the sink now logs a warning and drops the notification instead of sending it to `""`. It still moves the cursor back to the last valid block, which is what an undo does when no undo URL is set.
2. **Shutdown during retries:** if the context was cancelled, `send` and `recoverPending` now return the context error and skip the on-failure policy. The pending file is left alone, no termination message is written, and skip mode no longer deletes the saved payload. If the shutdown happens while a saved payload is being resent at startup, `Run` now returns `nil`, so the process exits cleanly instead of with 75.
3. **Cursor write failure after a successful POST:** the steps after a successful POST (saving the cursor, removing the pending file, updating the progress metric) now live in a new `commit()` function. It only logs its errors, as the code did before this PR. `deliver` now returns only errors from the HTTP call. `deliveryFailed` now uses a new `urlFor()` helper for its fallback URL, so a failed reorg notification reports the undo URL instead of the main webhook URL.

For the cursor write failure I chose "log a warning and continue" over stopping with a plain (non-75) error. If the disk write fails, the worst outcome is that the payload gets sent again after a restart.

sink/webhook/pending.go
```diff
@@ -0,0 +1,128 @@
...
+	"time"
+)
+
+// pendingDelivery is a payload that could not be delivered, kept on disk next
+// to the cursor file. It is written when every attempt has failed in
+// OnFailureExit mode and removed once the payload goes through, so the next
+// start delivers it before a Substreams stream is opened. A kill in the
+// middle of a call leaves no file: the cursor was not advanced, so the stream
+// re-sends that block.
+type pendingDelivery struct {
```
billettc: I don't see anything that limit the size of the pending stchuff. I feel that can be an issue when we are going to run it as a hosted service.

billettc (changes requested):

cmd/substreams/sink_webhook.go
```diff
@@ -24,20 +29,64 @@ func init() {
...
+	sinkWebhookCmd.Flags().Int("webhook-batch-max-blocks", 0, "Send up to this many blocks per call, in the batch payload shape (see below). 0 sends one block per call in the single-block shape")
+	sinkWebhookCmd.Flags().Duration("webhook-batch-max-wait", time.Second, "Longest a batch waits for more blocks before it is sent, checked when the next block arrives. A batch is also sent when the chain is live, before an undo notification, and when the stream ends")
+	sinkWebhookCmd.Flags().String("webhook-undo-url", "", "URL that receives a POST for each chain reorganization, with body {\"lastValidBlock\": {\"number\": ..., \"id\": \"...\"}, \"manifest\": {\"moduleName\": \"...\"}}. Empty disables the notification; the blocks that replace the undone ones are still delivered to <url>")
+	sinkWebhookCmd.Flags().String("webhook-termination-log", "/dev/termination-log", "File that receives the reason for a delivery-failure exit, written only when the file already exists (Kubernetes creates it)")
```
maoueh: Does the default really make sense?

sink/webhook/client.go
```diff
@@ -3,55 +3,179 @@ package webhook
...
+
+	var signingSecret []byte
+	if config.SigningSecret != "" {
+		signingSecret = []byte(config.SigningSecret)
+	}
+
 	return &Client{
 		httpClient: &http.Client{
 			Timeout: config.Timeout,
 		},
```
maoueh: We probably should ask Claude to tweak those based on the knowledge that we implement a webhook service.

sink/webhook/sink.go
```diff
@@ -2,38 +2,141 @@ package webhook
...
+const (
+	// OnFailureSkip logs the failure, drops the block and moves on to the next
+	// one. The cursor is not advanced past the dropped block, so a later
+	// restart replays it.
+	OnFailureSkip OnFailure = "skip"
+	// OnFailureExit keeps the block in the pending file, writes a termination
+	// message and stops the sink with ExitCodeDeliveryFailed. The next start
+	// delivers the pending block before it opens a Substreams stream.
+	OnFailureExit OnFailure = "exit"
+)
```
maoueh: Our best practices is to us `go-enum` for enums

sink/webhook/sink.go
```diff
@@ -2,38 +2,141 @@ package webhook
...
+	// TerminationLogPath receives the reason for a delivery-failure exit. It
+	// is written only when the file already exists, which is the case under
+	// Kubernetes, so the default of /dev/termination-log is safe elsewhere.
+	TerminationLogPath string
+	Logger             *zap.Logger
 }
 
 var WebhookCallsCounter = sink.Metrics.NewCounter("webhook_calls", "Number of calls made to the webhook")
 var WebhookSizeBytes = sink.Metrics.NewCounter("webhook_bytes_sent", "Number of bytes sent via webhook")
+var WebhookProgressBlock = sink.Metrics.NewGauge("substreams_sink_progress_block", "Last block number delivered to the webhook")
```
maoueh: Metrics are most probably prefixed already, looking at the two above.

sink/webhook/pending.go
sduchesneau: Pending is just a single webhook message. If we can send it as a webhook message, it won't be too big to store on disk. I don't think that's an issue...
What do you suggest?

sink/webhook/sink.go
```diff
@@ -82,19 +209,197 @@ func (s *Sink) Run(ctx context.Context) error {
...
+	}
+
+	if !pending.isUndo() && pending.Batched != s.batching() {
+		// The user switched batching on or off while the sink was down. The
+		// receiver expects the new shape, and the cursor was not advanced past
+		// these blocks, so the stream sends them again in that shape.
+		s.logger.Info("pending payload was written in the other batching mode, discarding it; the stream re-sends its blocks",
+			zap.Bool("pending_batched", pending.Batched), zap.Bool("batching", s.batching()), zap.Uint64("block", pending.BlockNumber))
+		return startCursor, removePending(s.pendingFile)
+	}
```
maoueh: Couldn't we de-construct the batch and send N messages instead? It says `the stream re-sends its blocks` but then what's the point of pending...

sink/webhook/sink.go
```diff
@@ -82,19 +209,197 @@ func (s *Sink) Run(ctx context.Context) error {
...
+
+	if err := writeTerminationMessage(s.terminationLog, failed.TerminationMessage()); err != nil {
+		s.logger.Warn("failed to write termination message", zap.String("path", s.terminationLog), zap.Error(err))
+	}
+	return failed
 }
 
+// writeTerminationMessage writes msg to path when path already exists. Under
+// Kubernetes the kubelet creates the file; anywhere else nothing is written.
+func writeTerminationMessage(path string, msg []byte) error {
```
maoueh: I don't like the `...Message` suffix, seems like a webhook message, let's find another term.

sink/webhook/sink.go
```diff
@@ -104,43 +409,130 @@ func (s *Sink) handleBlockScopedData(ctx context.Context, data *pbsubstreamsrpc.
...
 func (s *Sink) handleBlockUndoSignal(ctx context.Context, undoSignal *pbsubstreamsrpc.BlockUndoSignal, cursor *sink.Cursor) error {
-	// Save cursor to state file on undo
-	if s.stateFile != "" && cursor != nil {
-		if err := sink.WriteCursor(s.stateFile, cursor); err != nil {
-			s.logger.Warn("failed to save cursor to state file on undo",
-				zap.Error(err),
-				zap.String("file", s.stateFile))
+	// The receiver must see the blocks before it is told some of them are
+	// gone, so an open batch goes out first.
+	if s.batch != nil {
```
maoueh: **[bug]** `handleBlockUndoSignal` always flushes the entire open batch before applying the undo, including blocks with number > `lastValidBlock`. Those blocks have not been delivered yet. With `--webhook-undo-url` empty (the default), they are POSTed to the webhook and never revoked: the cursor is then written back to the last valid block and only replacements follow. That contradicts the changelog/help text that without an undo URL "only the replacement blocks are delivered". Example: cursor at 9, open batch `[10,11,12]`, undo to 9 → receiver gets 10–12 and then 10'–12' with no notification. `TestBatch_FlushedBeforeUndo` only covers the undo-URL path and expects this send-then-undo shape; there is no test for batching + empty undo URL.

**Suggestion:** Split the open batch at `lastValidBlock`: flush (or keep) the still-valid prefix, drop everything above it, then send the undo notification if configured. When the undo URL is empty, do not POST the discarded suffix. Add a test: batching on, no undo URL, in-flight batch straddling lastValid, assert the webhook never sees heights above lastValid.

sink/webhook/sink.go
```diff
@@ -104,43 +409,130 @@ func (s *Sink) handleBlockScopedData(ctx context.Context, data *pbsubstreamsrpc.
...
+		Fingerprint:    s.fingerprint,
+	}
+	if undoSignal.LastValidBlock != nil {
+		pending.BlockNumber = undoSignal.LastValidBlock.Number
+	}
+	if cursor != nil {
+		pending.Cursor = cursor.String()
+	}
+
+	return s.send(ctx, pending)
```
maoueh: **[bug]** A failed undo POST goes through `send`, which on `--webhook-on-failure=skip` (the default) logs and returns nil. The sinker then continues and delivers the replacement `BlockScopedData` messages. The receiver still has the fork and never got `{"lastValidBlock":...}`, so it cannot drop it before 11'/12' arrive. Exit mode is fine (pending undo, cursor stays, stream not opened). Skip mode breaks the invariant the feature is built on: "the receiver can drop the blocks above the last valid one before the replacements arrive." Same gap on recover: a pending undo that still fails in skip mode is deleted and the stream starts from the saved fork cursor (`recoverPending` returns `startCursor`).

**Suggestion:** Do not apply skip to undo payloads. Return the delivery error (or treat undo failures as exit) so replacements are not delivered until the notification succeeds. If skip must exist for undos, document that the receiver will see forks and replacements with no rollback. Add a test: undo URL 5xx, `OnFailureSkip`, then a replacement block, assert the replacement is not POSTed (or that the undo error stops the handler).

sink/webhook/client.go
```diff
@@ -3,55 +3,179 @@ package webhook
...
+		authHeaderName = DefaultAuthHeaderName
+	}
+
+	var signingSecret []byte
+	if config.SigningSecret != "" {
+		signingSecret = []byte(config.SigningSecret)
+	}
+
 	return &Client{
 		httpClient: &http.Client{
```
maoueh: **[suggestion]** `http.Client` uses the default redirect policy. Go strips `Authorization` on a host change, but not a custom `--webhook-auth-header-name` (e.g. `X-Api-Key`) and not `X-Substreams-Signature`. A 307/308 from the configured URL therefore forwards the secret header and a valid body signature to the redirect target. This is new attack surface from the auth/signature work; the URL is operator-supplied, but open redirects on that host are enough.

**Suggestion:** Set `CheckRedirect` to refuse redirects (typical for webhook senders), or strip `authHeaderName` and `SignatureHeader` on any host/scheme change. A test with a 307 to another `httptest.Server` should show the second request has neither header.

sink/webhook/sink.go
billettc: @maoueh is go-enum part of the `super power`?

PR #925, opened 2026-09-07
fix(cli): create config dir before writing registry token

GabrielCartier:
`substreams registry login` failed with a "no such file or directory" error when `~/.config/substreams` didn't exist yet, since the token write never created the parent directory. The write now creates the directory first, writes the token file with mode 0600 instead of 0644, and tightens permissions on an existing file during re-login. Adds a couple of unit tests and a changelog entry.


sduchesneau (approved):

sduchesneau:
just `go_bump google.golang.org/grpc@v1.83.1` before you can merge this @GabrielCartier

PR #931, opened 2026-09-10
chore: log per-store lines at debug

GabrielCartier:
Six log lines that fired on every store open and every squash for tier1/tier2 jobs are now Debug instead of Info: mmap/in-memory KV store selection, flushing store at boundary, merging partial into full store, deleting partial store, and squashing time metrics. These accounted for roughly 10-15% of substreams tier1/tier2 log bytes in dfuseio-global. Log-level only, no behavior change.

Note: `bin/test.sh` is already red on develop (`TestWithBlockBytesReadMeteringOptions`, `TestWithBytesReadMeteringOptionsGzip` in `metering/`), unrelated to this diff.

### DIFF
--- a/orchestrator/stage/squash.go
+++ b/orchestrator/stage/squash.go
@@ -152,13 +152,13 @@ func (s *Stages) singleSquash(stage *Stage, modState *StoreModuleState, mergeUni
 	modState.lastBlockInStore = rng.ExclusiveEndBlock
 	meter.mergeEnd = time.Now()
 
-	s.logger.Info("merged partial into full store",
+	s.logger.Debug("merged partial into full store",
 		zap.String("store", modState.name),
 		zap.Uint64("up_to_block", rng.ExclusiveEndBlock),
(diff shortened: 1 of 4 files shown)


orchestrator/stage/squash.go
```diff
@@ -182,7 +182,7 @@ func (s *Stages) singleSquash(stage *Stage, modState *StoreModuleState, mergeUni
 		}
 	}
 
-	s.logger.Info("squashing time metrics", meter.logFields()...)
+	s.logger.Debug("squashing time metrics", meter.logFields()...)
```
sduchesneau: keep just this one as Info.
Very useful when investigating performance issues, AND at least we will keep one out of the 6 to see squashing progress.

GabrielCartier: Done, `squashing time metrics` is back at Info. The other five stay at Debug, and the changelog now notes it is kept as the squash progress signal.

maoueh (approved):

PR #939, opened 2026-09-16
Disconnect clients when back-processing ends far behind

sduchesneau:
When a production-mode request back-processes for hours, the linear handoff block is fixed at the start and ends up far behind the chain, so tier1 processes the gap linearly. The same thing happens to a live request that cannot keep up with the chain: it drifts further behind as it goes.

Both are now disconnected with the usual `Unavailable` "endpoint is shutting down, please reconnect" error, using the same rule. Reconnecting from its cursor re-plans up to the new final block and back-processes the gap in parallel.

The rule compares a block with the last final block rounded down to a segment, capped by the stop block, and triggers when it is more than 2 segments behind. It is checked at two moments:

- right after back-processing, against the linear handoff block (`Tier1Service.blocks`);
- at every segment boundary of the linear phase, against the block that just became final (`Pipeline.handleStepFinal`). Only production-mode requests are checked, through the new `pipeline.WithFinalBlockLagCheck` option. The check runs once per segment, and detects the boundary by a change of segment rather than by a block number divisible by the segment size, since a chain can skip block numbers.

Neither disconnect quicksaves the stores: a quicksave at the cursor would make the reconnection load it, skip the parallel back-processing and process the gap linearly, which is what the disconnect avoids.

`SUBSTREAMS_MAX_LINEAR_HANDOFF_LAG_SEGMENTS` overrides the default of 2 segments for both checks.

`substreams gui` does not reconnect on its own yet; that is on the `gui-reconnect` branch.


sduchesneau:
@copilot review

maoueh (approved):
Approving, discussing offline about it.

PR #947, opened 2026-09-18
fix: support multiple foundational stores

driescroons:
## Problem

Modules can declare multiple foundational-store inputs, and the WASM runtimes assign each input its own index. The server only wired clients from the first input, so accessing index 1 failed with `invalid foundational store index 1, 1 stores declared`.

## Fix

Wire exactly one client for each foundational-store input in declaration order. Reject unsupported client cardinality before execution instead of silently corrupting guest indexes.

## Verification

- `go test ./pipeline/...`
- `go vet ./pipeline/...`
- `go build ./...`
- focused host index-routing tests
- release-note validation
- `git diff --check`


maoueh:
Hey, was this an agentic mistake or something else?

Happy to update skills/docs if it was, to
avoid this for someone else.

Unless it's something else …?

PR #949, opened 2026-09-21
Add remote store squashing with batch runs and local fallback

billettc:
Tier1 selects the store-merge implementation with a plugin DSN, the same shape as auth. Empty or `local://` keeps in-process squashing. `grpc://` is registered by the squasher project: each squash run is one RPC per store module, so empty-partial copies and merges happen off-box and tier1 does not read store files. If that remote is unreachable, the run is squashed locally after retries fail; a live remote that returns an application error still fails the request.

### DIFF
--- a/app/tier1.go
+++ b/app/tier1.go
@@ -26,6 +26,7 @@ import (
 	"github.com/streamingfast/substreams/pb/sf/substreams/rpc/v2/pbsubstreamsrpcv2connect"
 	"github.com/streamingfast/substreams/reqctx"
 	"github.com/streamingfast/substreams/service"
+	"github.com/streamingfast/substreams/squash"
 	"github.com/streamingfast/substreams/wasm"
 	_ "github.com/streamingfast/substreams/wasm/wasmtime"
 	"github.com/streamingfast/substreams/wasm/wazero"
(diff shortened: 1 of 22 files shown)


app/tier1.go
```diff
@@ -322,6 +328,13 @@ func (a *Tier1App) Run() error {
 		opts = append(opts, service.WithStoreSizeLimit(a.config.StoreSizeLimit))
 	}
 	opts = append(opts, service.WithExecOutPrefetch(a.config.ExecOutPrefetch))
+	squasher, err := squash.New(a.config.SquasherPlugin, a.logger)
```
sduchesneau: It might be clearer here if we ```if a.config.SquasherPlugin != "" {
...
append
}
```
so we don't have to follow squash.New to know that it will return nil for empty string, etc.

File: orchestrator/stage/squash.go [hunk 2]
```diff
@@ -354,16 +359,85 @@ func getPartialOrFullKV(ctx context.Context, modState *StoreModuleState, rng *bl
...
+		if !shouldFallbackToLocal(s.ctx, err) {
+			return nil, nil, err
+		}
+		s.logger.Warn("remote squasher not responding, falling back to local squash",
+			zap.Int("stage", stage.idx),
+			zap.Int("units", len(run)),
+			zap.Error(err),
+		)
+	}
+	return s.localSquashRun(stage, run)
```
sduchesneau: Every attempt to remote squasher, with the retries fibonacci sequence, will take between 10 and 20 seconds to fallback to local.

I'm not sure if we should:
1) add a mechanism that says "keep squashing locally for the next 5 minutes" so that we don't hit this VERY LONG fallback delay

or

2) disable fallback to local completely (which is what an operator might want)

but given that large jobs have hundreds and thousands of segments to squash, a fallback that takes 20 seconds to detect is not very effective

sduchesneau (changes requested):

sduchesneau (approved):

